PR #14, opened 2015-08-04
WEB-219 Implemented maintain-admin area including tests

sekrell:
This feature implements a view in the admin area, in which you can see the diskspace needed from system/programs/extracted programs/apks and backups. It's also possible to free the space through commands and web ui. 

Needed feature to gain disk-space on server for backup-functionality. Will implement/complete WEB-194 after successful merge.

Currently all phpspec-tests and behat-tests run through.

Changes were made according to your advices. Ready to merge.


MarkoBurazer:
yes, js should be in seperate files, as well as all css code

PR #19, opened 2015-12-07
Program download stats: circumvent proxy ip, add username to stats, use 'geo_plugin'

sjaindl:
- Uses HTTP_X_FORWARDED_FOR instead of REMOTE_ADDR to circumvent proxy, converts street ip's to regular ip's (as street ip's are not supported by geoip provider)
- Adds username to download statistics.
- Changes the geoip provider to 'geo_plugin' as it has a much larger geoip-database available.


src/Catrobat/AppBundle/Listeners/DownloadStatisticsListener.php
```diff
@@ -9,29 +9,61 @@
...
+
     public function onTerminateEvent(PostResponseEvent $event)
     {
         $attributes = $event->getRequest()->attributes;
         if ($attributes->has('download_statistics_program_id')) {
             $program_id = $attributes->get('download_statistics_program_id');
-            $ip = $event->getRequest()->server->get('REMOTE_ADDR');
-            $user_agent = $event->getRequest()->headers->get('User-Agent') ;
-            $this->createProgramDownloadStatistics($program_id, $ip, $user_agent);
+            $ip = $this->getIp($event);
```
chartinger: Why not just use [getClientIp from Request](http://api.symfony.com/2.7/Symfony/Component/HttpFoundation/Request.html#method_getClientIp)?

sjaindl: Haven't known that such a function already exists - I just call the function now!

PR #22, opened 2016-01-11
Remove oauth password

sjaindl:
Removes password replacement when OAuth user is merged with native user.
Login is done via session token instead.


sjaindl:
@chartinger @w00t2k1 : now everything is fixed & faked ;)
LDAP sign-in from Pocket Code works, too.

PR #108, opened 2017-01-24
Web 122 remixing information

rsamer:


rsamer:
Reorganised JS and CSS/Less files -> should be compatible with grunt.
Please upgrade to PHP v5.6

rsamer:
By the way: I've created a second folder for JS plugin files and renamed the existing.
The original folder was renamed to "globalPlugins" which contains those plugin files that are needed by every (!) HTML page. -> These files are concatenated with other globally used JS code and merged into the script.js file by Grunt.

The name of the new folder is "localPlugins" which contains only plugin files that are used by a few or a single HTML page. -> These plugin files are not concatenated and merged by Grunt, so the developer is responsible for including the file manually (-> same way as a developer would normally include a custom JS file) e.g.
`<script src=" {{  asset( getJavascriptPath('remixgraph.js') )}}"></script>`

PR #110, opened 2017-01-25
Quick fix Doctrine warnings

rsamer:
Fixed Doctrine inconsistency warnings (shown in Symfony web toolbar) as reported by Andrej and Eric.


rsamer:
Please clear symfony/doctrine cache (`php app/console cache:clear`) before you check the web toolbar in the browser. Otherwise the warnings probably won't disappear.

PR #123, opened 2017-02-20
WEB-328 no mail for username

selvex:
This PR implements a check if the username contains a `@` symbol and tells the user what the reason is why the registration failed.

It also adds error messages to tell the user when his entered passwords didn't match.


app/Resources/translations/catroweb.en.yml
```diff
@@ -466,6 +466,7 @@ errors:
     invalid: Your username is invalid
     exists: This username already exists.
     not_exists: This username does not exist.
+    not_email: The username must not be a email address.
```
selvex: Typo. Should be `an email address`

PR #137, opened 2017-04-10
WEB-392 refactor upload

selvex:
before: problem with upload = entry in db without files
now: problem with upload = no entry in db

refactor code to fit coding standard

refactor program upload and return failed upload response on failure

add symfony/filesystem to project, fix bug in upload controller begin uploading in tmp location

intermediate design save

WEB-391 refactor upload
before: error while uploading = entry in db without files
now: error while uploading = no entry in db without files

add directories


src/Catrobat/AppBundle/Entity/ProgramManager.php
```diff
@@ -1,391 +1,455 @@
...
     }
+    $program->setName($extracted_file->getName());
+    $program->setDescription($extracted_file->getDescription());
+    $program->setUser($request->getUser());
+    $program->setCatrobatVersion(1);
+    $program->setCatrobatVersionName($extracted_file->getApplicationVersion());
+    $program->setLanguageVersion($extracted_file->getLanguageVersion());
+    $program->setUploadIp($request->getIp());
+    $program->setFilesize($file->getSize());
+    $program->setVisible(true);
```
selvex: This has to be added here for WEB-376:
```php
    $version = $program->getCatrobatVersionName();
    $max_version = $this->getParameter('catrobat.max_version');

    if (version_compare($version, $max_version, ">"))
    {
      $program->setPrivate(true);
    }
    else
    {
      $program->setPrivate(false);
    }
```

selvex:
caused problems with composer
problems fixed in new branch, new pr created #149

PR #138, opened 2017-04-11
WEB-375_PenalizeComments bann user for comments

selvex:
adds the ablity to bann a user from reported comments

add migration and update user entity

add function to ban user, ban user and delete comment after pressing button

add option to enter a reason for bann, add stub for sending mail

change creation of messages

set user to locked, remove reason, add static reason

add user checker

inject doctrine into user checker

automatically unban user after time

make user manually bannable again

add tests for banned users

add testing if user was created in db

finalize WEB-375 as discussed with s schrimpf

defuse migration bomb

### DIFF
--- a/app/DoctrineMigrations/Version20170221152814.php
+++ b/app/DoctrineMigrations/Version20170221152814.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Application\Migrations;
+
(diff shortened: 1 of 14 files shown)


selvex:
The backend stuff is complete but the front end parts were edited in files that are not commited to git.

PR #148, opened 2017-04-28
WEB-382 restrict unallowed searchtexts

Richard014:
WEB-382 removed filter from js code and added filter to php code


src/Catrobat/AppBundle/Controller/Api/SearchController.php
```diff
@@ -21,6 +21,12 @@ public function searchProgramsAction(Request $request)
   {
     $program_manager = $this->get('programmanager');
     $query = $request->query->get('q');
+
+    $query = str_replace("yahoo", "", $query);
+    $query = str_replace("gmail", "", $query);
+    $query = str_replace("gmx", "", $query);
+    $query = trim($query);
```
selvex, line 28: There would have been a way to avoid triming the query, but I don't expect it to cause any troubles so we will accept this solution.

src/Catrobat/AppBundle/Features/Web/search.feature
```diff
@@ -16,4 +16,16 @@ Feature: Searching for programs
...
+    Given I am on "/pocketcode/search/yahoo%20myprog"
+    Then I should see "Your search returned 2 results"
+
+  Scenario: search for gmail should search for all
+    Given I am on "/pocketcode/search/gmail"
+    Then I should see "Your search returned 4 results"
+
+  Scenario: search for gmx should search for all
+    Given I am on "/pocketcode/search/gmx.at"
+    Then I should see "Your search returned 0 results"
```
selvex, line 31: Tests had to be fixed since our website changed entirely.

selvex (approved):
Works.

PR #162, opened 2017-11-29
Pre Move Merge

HCrane:
Merges current dev-master into master to keep master up to date before we move to new server!


HCrane:
Since the deployment approaches we will merge the develop into the master one more time. The next merge will hopefully be the release of Catroweb 2.0

PR #163, opened 2018-01-22
WEB-446_Girl_Flavor

HCrane:
Makes Girl Flavor releaseable. Not release ready yet some infos are missing.


web/css/themes/makeITup/catro_bootstrap.less
```diff
@@ -0,0 +1,323 @@
...
+
+          span
+          {
+            margin-right: 10px;
+          }
+        }
+      }
+    }
+  }
+}
```
selvex, line 323: I believe there is a way to import the base theme and then overwrite the needed things.
This looks like you copied a whole file over which is something you should never do. Just overwrite the things that should be different.

selvex (changes requested):
Generally a lot of file copying. I will check if this couldn't be done by importing stuff (for the html side)

The css side should definitely be done using `@import`.

web/css/themes/makeITup/catro_bootstrap.less
selvex: As we discussed, this is really an issue. I just ran into the situation that I would have to copy a change into by now 7 different files.

That's not how we initially intended the `catro_bootstrap.less` to be used.

We will solve this issue inside this pr.

PR #164, opened 2018-02-06
Web 413 RTL Support

HCrane:
Adds RTL Language Support
Known problems:

- Numbers & Dates might be malformated
- Spacing with Icons could not be according to standard

### DIFF
--- a/app/Resources/themes/phirocode/index.html.twig
+++ b/app/Resources/themes/phirocode/index.html.twig
@@ -17,7 +17,7 @@
 
   <div id="newest">
     <h1>{{ "programs.newest"|trans({}, "catroweb") }}</h1>
-    <div class="programs"></div>
+    <div class="programs" dir="ltr"></div>
   </div>
 
   <div id="recommended">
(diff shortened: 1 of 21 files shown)


app/Resources/themes/phirocode/index.html.twig
```diff
@@ -34,22 +34,22 @@
              width="30" height="30" />
       {% endif %}
     </div>
-    <div class="programs"></div>
+    <div class="programs" dir="ltr"></div>
```
selvex, line 37: After adding `dir="ltr` to the html tag it is set for the *whole* page [1]. There should not be a need to set it to indiviual pieces of the website.

[1] https://www.w3.org/International/questions/qa-html-dir

selvex (commented):
We need to check if it is ok to just set the html tag to `dir="ltr"` or if we have to apply it to individual pieces.

app/Resources/themes/phirocode/index.html.twig
HCrane: That is correct yes. Problem is this should always be ltr wether it is Arabic or English. That is the reason why i hardcoded it in there as well as in some other cases like Code Statistics. It would also be possible to set it dynamically but it was my understanding that this should always be set to ltr.

>Below the html tag, only use the dir attribute on structural elements where the base direction needs to change in order for the text to display correctly.

This is from your [1] so basically we need it in there but wether it is dynamic or not is up to us.

selvex: As I understand it this would cause program names and other information regarding the program (f.e. time uploaded now, more information after the redesign) being "mirrored" for users of the RTL page.

Is that really expected behaviour?

app/Resources/views/Default/base.html.twig
```diff
@@ -23,7 +23,7 @@
   {% block head %}{% endblock %}
 </head>
 <body>
-
+{{ app.request.locale }}
```
HCrane, line 26: This should be removed since it only was for debug output.

HCrane:
We will put this on hold untill we have completely switched to bootstrap.

PR #165, opened 2018-02-13
WEB-456 Simple notification on comment

selvex:
this adds the functionality to notify a user when another user commented on ones program
it leverages the generic notification system from WEB-403

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -65,6 +65,12 @@ security:
             simple_preauth:
                 authenticator: apikey_authenticator
 
+        api_check_notifications:
+            pattern: ^.*/api/notifications/get.json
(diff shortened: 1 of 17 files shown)


app/config/security.yml
```diff
@@ -65,6 +65,12 @@ security:
             simple_preauth:
                 authenticator: apikey_authenticator
 
+        api_check_notifications:
+            pattern: ^.*/api/notifications/get.json
+            stateless: true
+            simple_preauth:
+                authenticator: apikey_authenticator
+
```
selvex, line 73: We may want to use symfonys apikey system to log the user in during the checks for notifications, since the app probably can't deal with session cookies.

PR #167, opened 2018-02-19
WEB-394 Cant download program

HCrane:
Makes the correct cleanup of Temporary files that get created during upload. Previously if the upload was interrupted (whatever cause) those files persisted and an entry was wirtten into the database. This PR should fix that.
Also 2 new commands get implemented; one to set all programs that got affected by this to invisible and another command that makes it possible to set visible.

### DIFF
--- a/src/Catrobat/AppBundle/Commands/InvalidFileUploadCleanupCommand.php
+++ b/src/Catrobat/AppBundle/Commands/InvalidFileUploadCleanupCommand.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Catrobat\AppBundle\Commands;
+
+use Catrobat\AppBundle\Entity\Program;
+use Catrobat\AppBundle\Entity\ProgramRepository;
+use Doctrine\ORM\EntityManager;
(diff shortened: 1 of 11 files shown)


src/Catrobat/AppBundle/Commands/InvalidFileUploadCleanupCommand.php
```diff
@@ -0,0 +1,59 @@
...
+    $ids = explode(",\n", $content);
+
+    /** @var ProgramRepository $pr */
+    $pr = $this->getContainer()->get('programrepository');
+
+    $fs = new Filesystem();
+    /** @var EntityManager $em */
+    $em = $this->getContainer()->get('doctrine.orm.entity_manager');
+
+    foreach ($ids as $id)
```
selvex, line 47: We have to check if the program with the given id exists in the database and skip it if not. Otherwise this would cancel the command if there are any program ids that don't exist before setting them to invisible. If we don't add this check the command might stop mid-execution and only half of the programs would be invisible, that would not be good.

selvex (changes requested):
Add check for null pointer exception (program not in database) during command execution.

selvex (approved):

PR #168, opened 2018-02-23
WEB-446_Girl_Flavor

HCrane:
Renames old lunacat flavor and makes it release ready. Also some
small changes have been to correct an visible issue with the help-text
next to the recommended programs.

There is a temporary file that should not be uploaded but apperantly gitignore is ignored?


web/css/luna/licensetoplay.css
```diff
@@ -0,0 +1 @@
+article>div{color:#8a78a8}
```
selvex: There is no article on the page anymore. This css affects nothing, we should find the less line that causes this to be generated and correct it.

web/css/luna/termsofuse.css
```diff
@@ -0,0 +1 @@
+article>div{color:#8a78a8}
```
selvex: Same as above

src/Catrobat/AppBundle/Commands/invisible_programs/executed/2018-02-19_16_29_34
```diff
@@ -0,0 +1,175 @@
+12821,
+17146,
```
selvex: Gitignore problem

selvex (changes requested):

selvex (approved):

PR #171, opened 2018-03-04
Updated README.md

Ash-Kay:
added space between ### and How, Now styling works properly!
'M' of made is changed to lowercase.


selvex:
Thank you. We will add this change to the dev-master.
I will close this pull request since the source branch is `master` but it should be `dev-master` and will add this change in the next merge.

PR #175, opened 2018-03-28
WEB-415 welcome section

Richard014:


app/Resources/views/index.html.twig
```diff
@@ -7,6 +7,40 @@
 
 {% block body %}
 
+  {% if not app.user and not isWebview() %}
+    <div id="welcomeSection">
+      <h3>{{ "programs.welcome.title"|trans({}, "catroweb") }}</h3>
+      <h3>{{ "programs.welcome.subtitle"|trans({}, "catroweb") }}</h3>
+      <div class="row">
+        <div class="col-md-8">
+          <img src="{{ asset('images/help/starters.png') }}" height="200px" width="33%" onclick="self.location.href='{{ url('catrobat_web_starter') }}'" />
```
selvex: Don't use `width: 33%`. We would like to have it dynamic which means you should use another bootstrap grid inside here. Just start a new row and use `col-md-*` again. (nested rows work)

Don't use onclick listeners. The mouse pointer won't change which is bad usability. Use an anchor that contains the image tag and set the href of the anchor and you are golden.

Instead of the images which get distorted I would love to see you make "cards" using css which look the same as the images you used. It shouldn't be that hard using bootstrap. It would take a while but would be worth the effort.
If you prefer not to, this is ok for now. 
The 2 other changes are mandatory.

app/Resources/views/index.html.twig
```diff
@@ -7,6 +7,40 @@
...
+          <img src="{{ asset('images/help/tutorials.png') }}" height="200px" width="33%" onclick="self.location.href='{{ url('catrobat_web_tutorialcards') }}'" />
+          <img src="{{ asset('images/help/step_by_step_mobile.png') }}" height="200px" width="33%" onclick="self.location.href='{{ url('catrobat_web_stepByStep') }}'" />
+        </div>
+        <div class="col-md-4">
+            <iframe id="youtube-help-video" height="200px" width="100%" src="https://www.youtube.com/embed/BHe2r2WU-T8" frameborder="0" allowfullscreen></iframe>
+        </div>
+      </div>
+
+      <div class="row">
+        <div class="col-md-6 d-block inline">
```
selvex: The classes `d-block` and `inline` don't exist inside our css and as far as I could find neither do they in any js library. You can savely remove them if you don't know the purpose of them.

app/Resources/views/index.html.twig
```diff
@@ -7,6 +7,40 @@
...
+          {# todo add link to new statistic page #}
+          {% set stats = getCommunityStats()  %}
+          {{ stats['program_count'] | number_format(0, ",", ".") }} {{ "footer.programs"|trans({}, "catroweb") }}
+          {# <a href="{{ path('homepage_click_stats') }}">{{ "programs.welcome.statistics"|trans({}, "catroweb") }}</a> #}
+        </div>
+      </div>
+  </div>
+  {% endif %}
+
+  <div id="test123">
```
selvex: Pretty sure you can remove this.

app/Resources/views/index.html.twig
selvex: This holds for any occurences of `d-block` and `inline` classes inside this PR

selvex (changes requested):

app/Resources/views/index.html.twig
Richard014: The css for bootstrap cards seems not to be included in our project.
I tried to use a link to a bootstrap stylesheet, but the cards didn't looked nice.

src/Catrobat/AppBundle/Features/Web/Context/FeatureContext.php
```diff
@@ -237,6 +237,40 @@ public function iShouldSeeTheFeaturedSlider()
...
+  }
+
+
+  /**
+   * @Then /^I should not see the welcome section$/
+   */
+  public function iShouldNotSeeTheWelcomeSection()
+  {
+    assertNull($this->getSession()->getPage()->findById('welcomeSection'));
+  }
```
selvex: These functions are duplicates.

selvex (changes requested):
Good.
I will apply the final changes myself during code acceptance.

selvex (approved):

selvex:
Transfered changes to all flavor files, fixed click statistic tests.

PR #177, opened 2018-03-29
WEB-461 change report button & regenerate grunt files

Richard014:


selvex (approved):
Ok for now, `display: table` is not optimal but we will rewrite the enitre page soonish anyway.

PR #185, opened 2018-07-09
WEB-478_Deletion_footer_statistics

HCrane:
It was required by our PO's to remove all statistics from the footers and the welcome section. It was already done by a quick'n dirty fix on live server because of urgency.


selvex (approved):
ok

PR #187, opened 2018-07-10
WEB-488 Resolve Routing deprication

HCrane:
This resolves all routing deprications according to https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md#routing.
`UrlGeneratorInterface::ABSOLUTE_URL` is the default option. In case something else was used it got changed accordingly. Also all files touchd got code reformatted.


selvex (approved):
ok

PR #190, opened 2018-08-01
WEB-498 Resolve Console deprication

HCrane:
Resolves Console deprications in self written commands.


dmetzner (approved):
checked with tests

PR #191, opened 2018-08-01
Web-504 Resolve FrameworkBundle deprications

HCrane:
Only this line had to be edited all other stuff was not used.


dmetzner (approved):
checked with tests

PR #196, opened 2018-08-07
Web 451 console command delete apks

dmetzner:
This was done to make more space available on our hard drives and get rid
    of old APK files, but still keep the newest APKs available to the users.
    
    Example call:
     >php bin/console catrobat:clean:old-apk 3
    
This would delete all APKs older than three days. The age of APKs depends
    on the time of their last access. (Unix atime)
    
This command depends on the fact that APKs are saved with their ID
    as filename to work correctly.

### DIFF
--- a/app/config/commands.yml
+++ b/app/config/commands.yml
@@ -3,6 +3,10 @@ services:
               class: Symfony\Component\Filesystem\Filesystem
               public: false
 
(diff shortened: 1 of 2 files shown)


HCrane (dismissed):
Does work as intended. Maybe later we could add some logging but i dont think it is neccesary.

PR #197, opened 2018-08-07
Web 445 increase description maximal length

dmetzner:
Currently, the description is limited to 1000 characters when uploading. Since text is cheap the limit is now increased to 5000 characters.


dmetzner:
Was done with WEB-441 and is no longer hardcoded -> uses a parameter in services.yml (catrobat.max_description_upload_size)

PR #204, opened 2018-08-29
WEB-395 - ProgramLoader rework (search incorrect)

dmetzner:
- Translations were missing: since using the symphony translator or twig
  variables in external js files is not possible without the rewriting of
  many files (all that use the “ProgramLoader”). That’s why there is now
  a new Controller “TranslationController”. This controller can be called
  with “Symfonys” routing system and returns the requested translations.
- Added a new translation for no programs
- Coding standard applied
- Renamed many variables to more meaningful names.
- Reduced code redundancy
- Session System is now also in use for the Search result page when the
  same query gets executed or the page gets refreshed.
- Fixed brain.scss for bootstrap 4 -> Now programs are not overlapping
  on some screen sizes.
- The calculations for how many programs to show/load are improved.
  Even with resizes all rows are going to be filled when there are
  enough programs.
- Show more/less button functionality improved for a cleaner design.
  (full rows)
- Show more/less buttons are now only displayed when they should be.
- Searching didn’t work as expected because the database query used left-
  joins. With Left-joins using “setFirstResults” and “setMaxResults”
  the query result is not what you would expect (see documentation).
  Now a paginator handles the query.
- Since the testing required many more programs in the database than
  20, the Reset Command now has another option (“—more”). This will
  download around 200 random programs.


HCrane (approved):
Approved as discussed.

PR #205, opened 2018-08-29
WEB-524 Add PHPCheckstyle to project

HCrane:
adds a new Method for us to check if our developers
did work according to our coding standard. Also a
good tool to see what has to be done so we have clean code.


dmetzner (commented):
- Works as intended.
- Tests show no errors

PR #206, opened 2018-08-30
WEB 473 Set Logging with Monolog to rotate logs

dmetzner:
* changed monolog config to rotate logs
  - 1 new file every day
  - auto remove after 30 days
  - different log files are now in seperate folders

*  Logs Controller refactored + file search depth increased for the folders

* logs.html replace regex fixed for folders


HCrane (approved):
Approved.

PR #207, opened 2018-08-30
Web 526 fixing colors login register page

dmetzner:
fixed colors and logos of login and register page to match themes


HCrane (approved):
Approved as talked with Daniel

PR #216, opened 2018-09-10
WEB-534 Comment Refactoring

dmetzner:
The comment sections on a program page still used old alerts
  hardcoded text, had many code redundancies and also had some bugs.

- comments orderd newest first
- only 5 comments a time visible
- when there are more comments for a program the same show more/less
  buttons functionality was added like for programs (proramloader)
- a comment can get reported (if not your own program)
- a comment can get deleted by the author or an admin
- report and delete buttons use now fontawesome icons
- icons increase in size when hovering over them
- the comment image now show the authors avatar if he has one
- comments got their own html.twig file with reduced code
- also the comments now have their logic in their own js file
- commentControler now return statusCodes instead of hardcoded strings
- report/delete action requires a confirmation now (sweet alert)
- report/delete buttons only visible when user is allowed to perform
  those actions
- translations added
- statusCodes added
- error alerts replaced by sweetalert
- buttons icons use tooltips
- included session history: when user clicks show more several times
  and refreshes the page, the same amount of comments will be visible
- now a user will not get notified when he comments on his own programs
- added many new test scenarios


HCrane (approved):
1. Checkstlye :heavy_check_mark:
2. Code :heavy_check_mark:
3. Tests :heavy_check_mark:

Ready for merge!

PR #221, opened 2018-09-19
WEB 545 Report program rework

dmetzner:
- report container replaced by multi-input sweetalert2
 - pop up: when "send report" is pressed and no valid inputs are given
   the popup reopens and shows the errors/ what's missing
 - the report feature now has a session storage, so if a user writes a report and presses cancel
   when he opens the pop up again it's already filled with his data. (or on a page reload)
 - when reported the user gets a success message and is forwarded to the homepage
 - or if something goes wrong an error message
 - when a user is not logged in and wants to report a program he is forwarded to the login page
 - a user won't see the report button for his own programs
 - report button design changed similar to the link button
 - created custom pocketcode radio buttons (primary color design instead of OS/browser design)
 - report logic in ProgramReport.js
 - updated translations
 - fixed some code style errors & warnings
 - updated and wrote new report tests
 - also, I wrote some login and report API tests!

**On the Website we want a user to be signed in bevor reporting, the API, on the other hand, allows reporting a program without being signed in!**


src/Catrobat/AppBundle/Controller/Api/SecurityController.php
```diff
@@ -249,12 +249,14 @@ public function loginNativeUser(Request $request)
...
-        $correct_pass = $userManager->isPasswordValid($user, $password);
+        $factory = $this->get('security.encoder_factory');
+        $encoder = $factory->getEncoder($user);
+        $correct_pass = $userManager->isPasswordValid($user, $password, $encoder);
+        $dd = null;
         if ($correct_pass)
         {
           $retArray['statusCode'] = StatusCode::OK;
           $user->setUploadToken($tokenGenerator->generateToken());
-          $retArray['token'] = $user->getUploadToken();
```
HCrane: This should be in the new commit.

src/Catrobat/AppBundle/Features/Api/login_or_register.feature
```diff
@@ -1,13 +1,13 @@
 @api
```
HCrane, line 1: This feature is depricated since LoginOrRegister is depricated and replaced by 2 other functions. This will be deployed when the time is right for android and ios and also xwiki.

HCrane (approved):
Works like a charm!

| Status | Review Item |

src/Catrobat/AppBundle/Controller/Api/SecurityController.php
dmetzner: k, but this will result in failing tests till the API gets fixed

PR #223, opened 2018-09-20
WEB-546 remove bootstrap template page

dmetzner:
Removed the bootstrap template page. If information about bootstrap is needed use their homepage or google.


HCrane (approved):
approved

PR #222, opened 2018-09-19
WEB 541 Fontawesome

dmetzner:
- replaced all images that could have been replaced by fontawesome,
- updated design to fit fontawesome instead of image (the look should be close to the old one)
- added additional design changes to fit better for fontawesome (like buttons, etc)
- fixed minor css issues for themes and removed tables (now css alignment)

Now everything should be in the correct theme colors. (except default avatar image)

- Added icons to to the profile and myprofile page ;)

PR #224, opened 2018-09-20
WEB-547 Themes fix

dmetzner:
- removed deprected theme files
- updated index.html with twig conditions to fit all themes

- removed code redundancy from themes
- fixed / updated imports in scss files
- removed scss warnings
- fixed galaxy and alice theme

How to for themes:

  - overwrite variables in _theme_settings.scss if needed
  - if you need to customize css and there is no variable for it,
    replace the hardcoded css values with variables and include the
    variables in _variables.scss
  - if a page has different elements, like buttons, or a different Logic (js, route)
    behind it use twig conditions instead of coping files and changing minor things


web/font_awesome_wrapper/fontawesome-free/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "_args": [
     [
-      "@fortawesome/fontawesome-free",
-      "/home/catroweb/Catroweb-Symfony"
+      "@fortawesome/fontawesome-free@^5.3.1",
+      "/home/heroiceraser/Catroweb-Symfony"
```
HCrane: As already said in WEB-541(PR: #222)

HCrane (approved):
Works wonderfully. package.json will be fixed in #222 

| Status | Review Item |

PR #226, opened 2018-09-25
WEB-548 Display program tags and extensions using bootstrap badges

mfuerni:
Change style of tags and extensions by using bootstrap badges.
Redo structure and styling using CSS flexbox.
Adapt test cases and use XPath


web/css/custom/program.scss
```diff
@@ -1,4 +1,7 @@
+@import '../../bootstrap_vendor/scss/functions';
```
HCrane, line 1: Please import these files in the appropriate files in the base dirctory.

web/css/custom/program.scss
```diff
@@ -735,7 +739,39 @@ input[type='number']
...
+  }
+
+  #tags, #extensions
+  {
+    display: flex;
+    flex-direction: row;
+    align-items: baseline;
+    margin-top: -0.15rem;
+
+    @include media-breakpoint-down(sm)
```
HCrane, line 682: This is the new bootstrap approach to deal with different devices which is quite nice. Please since you are the first one to use it, switch all other statements to this new one.

HCrane (changes requested):
| Status | Review Item |

Basically ready for merge but if we introduce something new change also the old attributes to the new ones.

HCrane (approved):

PR #228, opened 2018-10-24
Move Google Login to auth2 api and fix infinite blocked popups

ckissane:
Attempts to fix the infinite blocked popups issue when trying to login through google.


ckissane:
This was already fixed in the php7 branch so I closed this.

PR #231, opened 2018-10-25
WEB-559 #selector jumps below title

fladens:
On the homepage when adding the #selector to the URL it should jump
exactly to the top of the selector title.

Added a :target css to index.scss so that the title has a top offset
Also added a scroll to the title to the Main.js, in case that the
content loads after the jump.

### DIFF
--- a/app/Resources/views/Default/baseParent.html.twig
+++ b/app/Resources/views/Default/baseParent.html.twig
@@ -52,12 +52,12 @@
 {% endblock %}
 
 <script>
-  let search_input = $('.input-search')
+  let search_input = $('.input-search');
 
   search_input.tooltip({
     trigger  : 'manual',
     placement: 'bottom'
-  })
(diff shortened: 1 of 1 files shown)


dmetzner:
I tested the implementation for following link: **pocketcode/#mostDownloaded**

- I think the desired representation would be the one that can be seen here
correct
 I only get this view if i re-enter the link.

- When i refresh a page or load it for the first time i get following representation:
incorrect
 Seems a bit **inconsistent**?

- Last but not least, please use **variables instead of hardcoded values**. This allows an easy way to adjust things for differerent themes. not all headers have the same size. (E.g. create@school)
theme

fladens:
@dmetzner 
I changed it to javascript function so it waits for the page to be loaded and then it calculates the height and jumps to the title.
Added load, hashchange and immediate jump.

PR #243, opened 2019-02-13
WEB 583 Mail System

dmetzner:
**parameters.yml and swiftmailer config have to be configured when we get an catrobat mail!** 

Fixed/Recreated the mailing system and updated it in a way that we now send html emails 
and plain text just as back up.

**Mails get send:** 
- account confirmation on registration
-  resetting the password 
- when triggered by the admin tool 'send mail'

**Changes:**
- updated config and admin config to work with mail system

- created/fixed: check email html.twig pages
- created mail templates
- fixed password length validation (change/reset)
- updated reset password page to fit the new design 
     -  show/hide password toggler added
     -  error messages fixed
   
- admin tool send email reworked

- Additional changes:
  -- login theme updated for large screen
  -- privacy policy design updated for large screens


dmetzner:
updated template with "no reply" message

HCrane (approved):
Everything is okay. We just have to get the credentials forthe mail server used, then this can be merged

dmetzner:
done in SHARE-15 (flex)

PR #248, opened 2019-02-21
SHARE-8 - Cleanup 

dmetzner:
-  Social Sharing stuff removed
- PhpDoc Blocks in 90% of all "src" files
 - bugs/warning fixes accross all files
- some deprecations fixed (Metadata, ..)
- views structure refactored
- composer updated with missing packages (ext-zip, ext-pcntl, ext-mbstring)
 - splitted multiple classes in one file into 1 class per file (symfony-flex will need this


HCrane (approved):
No tests were destroyed. This should help us in the future quite a lot.

PR #251, opened 2019-03-11
SHARE-3 Tutorialcards Redesign

mfuerni:
- Fix broken Tutorial-Cards design
- Add lang-Attribute to html-Tag (for hyphenation)

Maybe use larger images later, but enough for release


dmetzner:
can be merged 

- tests passed
- looks good (could have bigger images in the future, but as discussed enough for now)

PR #253, opened 2019-03-12
SHARE-22 remove old sweetalert usage

dmetzner:
sweetalert is already in the base script, no need to include it in every file


HCrane (approved):

HCrane:
Works as intended

PR #260, opened 2019-03-13
SHARE-27 No email as username

HCrane:
- fixes regex that takes care of basic validation of username
- adds bounds (username may be between 3 and 180 chars)
- adds test for it


dmetzner (changes requested):
also for the fosuser web registration

dmetzner (approved):
done for fos in SHARE-19

PR #265, opened 2019-03-16
SHARE-29 App version

dmetzner:
- now there is an app-version on every page (but not visible to the user) baseparent.html.twig
- added app-version to env
- added twig extension to use env variables
- added behat tests and new feature context functions


HCrane (approved):
does what it should. merge after backup tests are complete

PR #266, opened 2019-03-21
SHARE-32 backup fixed

dmetzner:
- timeout increased to 1day
- purge uses migrations now


HCrane (approved):
Does as it should and does not break anything.

PR #274, opened 2019-04-16
SHARE-46 remove recommended programs section

marweb1996:
Only show recommended programs section on main page if there are some available programs. Otherwise this section will be hidden.


dmetzner (changes requested):
tests/behat/features/web/recsys.feature:30 broken

dmetzner (approved):
✓

PR #276, opened 2019-04-17
SHARE-47 Media-Library: Changes Download Filename

mfuerni:
Changes the file name of downloaded assets in the media library to the name of the file instead of the id.
Adapts test case to test for correct file name.


dmetzner (approved):
✓

PR #280, opened 2019-04-22
SHARE-57 sidebar notification badge display bug

marweb1996:
The notification count badge in the sidebar should only be displayed if there are unread notifications.
Now the badge has display:none as default value which will be changed only if notification count > 0.


dmetzner (approved):
✔

PR #282, opened 2019-04-28
SHARE-56 Media-Library: New Ajax Loading + Separate Theme Special Category

mfuerni:
First, it completely changes how files are loaded via ajax
in the media library.
Previously, for every category there was a own API call.
Now there is a single API call, which takes the package into account.

Second, it changes the handling of the theme-special category.
Previously, every file tagged with a flavor was loaded into the
theme-special category. Now those files are sorted to the beginning
inside their category, and a new "ThemeSpecial" category is introduced.

Third, it fixes old test cases and inserts new ones for new behaviour.


src/Catrobat/Controller/Web/MediaPackageController.php
```diff
@@ -108,7 +108,7 @@ public function MediaPackageAction(Request $request, $package_name, $flavor = 'p
 
     $categories = [];
 
-    if ($flavor !== "pocketcode")
+    if ($flavor !== "pocketcode") // TODO: should we show theme-special also for Pocketcode?
```
mfuerni: @HCrane Please ask PO if we should hide ThemeSpecial for PocketCode flavor

mfuerni: Clarified. Hiding category for pocketcode is expected behaviour.

dmetzner (approved):
✔

PR #283, opened 2019-05-02
SHARE-51 introducing UUIDs

dmetzner:
The program.id and user.id have been updated from using integers
to using UUIDs. This means the user now sees different links
when beeing on a program/profile page.

- A new migration was added which updates all foreign key constraints
and all entities which use the program or user id.

- some entites have been updaterd too work with uuids

- Routing annotions have been changed since they accept now strings and not ints

- some logic changes files have been applied to work with strings and not ints

- the behat feature and context files have been updated to work with fixed ids.


dmetzner:
before deploying this PR, we have to create a backup.
Then deploy and remove the old database.
Then run all migrations and import the backup!

dmetzner:
merge after 112

HCrane (approved):

PR #284, opened 2019-05-02
SHARE-59 Uploading a program with followers fix

dmetzner:
**SHARE-59 Uploading a program with followers fix**

_What was the problem?_
The UploadController is not extending the Default Controller,
therefore we can't access services via 'this->get(...)'.

_How was it fixed?_
Now the catroNotificationService is a private member of the UploadController
and is configured via autowire like it is done for all the other members.


HCrane (approved):
✔️

PR #285, opened 2019-05-08
SHARE-63 Media-Library: Labels for File Icons + Sound Preview

mfuerni:
First, it shows the name over every file icon (if there is no thumbnail)
such that the user can distinguish non-image files.

Second, it allows to play sounds without downloading the files.


dmetzner (approved):
✔️

PR #286, opened 2019-05-09
SHARE-41 Delete Logs

mfuerni:
Fixes the clean logs command and the clean logs tool in admin backend.
It deletes everything inside the log folder.

Includes a PHPUnit Test.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -630,6 +630,11 @@ services:
     tags:
       - { name: console.command }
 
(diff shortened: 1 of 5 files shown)


dmetzner:
works as expected ✓

A small remark to the unit test: The depreciation warnings are caused by the Sonata Bundle and not by us.

dmetzner (approved):

PR #287, opened 2019-05-09
SHARE-64 Limit Text to Display Width

mfuerni:
Prevent text (especially headings) from overflowing the page border.
At first, try to reduce font size.
If this does not help, force breaks between words.

Refactoring:
 - Remove old jquery-2.1.0.js from globalPlugins.
 - Remove JS Standard Style NPM package


dmetzner (approved):
✓

PR #288, opened 2019-05-12
SHARE-67 Translations instead of static text

dmetzner:
- Replaced all found hardcoded messages with translations
- Replaced some split messages (part1, 2,..) with paramterized translations.
- Admin interface uses only english hardcoded messages!

- refactored notification system to use translations.
  ~ Message and title is only needed in the template, not in every object allocation.
  ~ Reduced duplicate code via extending a base notification
  ~ LikeNotification is now also triggered for (wow, love, smile)


templates/Default/footer.html.twig
```diff
@@ -69,8 +69,8 @@
         </select>
       </div>
       <div class="copyright footer-right">
-        Copyright &copy; {{ "now"|date("Y") }} <a href="http://www.catrobat.org">Catrobat</a>. All
-        rights reserved.
+        {{ "copyright" | trans({}, "catroweb") }} &copy;
```
mfuerni: I think it would be better to use a single translation with the year as variable.

HCrane: But then we have to delete the Translation every year..

mfuerni: Not if you use the year as variable/parameter/argument.
See second point of commit message.

dmetzner:
using a variable would not mean we have to change it every year. I just thought we may could need just the word "Copyright" one time. But changing that is no problem ;)

mfuerni:
> 
> 
> using a variable would not mean we have to change it every year. I just thought we may could need just the word "Copyright" one time. But changing that is no problem ;)

But maybe the year or the copyright-symbol is on another position in other languages. I think that it would be easier for translators if they have the full phrase. Maybe we should use 'footer-text' or 'copyright-notice' as keyword.

dmetzner:
I mean those were two separate sentences. But yes you are right. I will never know all the languages. So to be sure everything works as intended I changed it.

templates/Default/footer.html.twig
HCrane: Point taken :)

HCrane (approved):
:heavy_check_mark:

HCrane:
Please also remind me to update translations in a few days.

PR #289, opened 2019-05-12
SHARE-72 Password toggler for login page

dmetzner:
The password toggler for the login page was added.
Also a behat test for it  was added.


HCrane (approved):
✔️

PR #290, opened 2019-05-12
SHARE-71 Profile program loader rework

dmetzner:
- profile loads now all programs at once
- profile loads programs now orderd (newest first)
- program loader was fixed to show all programs even if rows are not full
- added behat tests


HCrane (approved):
✔️

PR #291, opened 2019-05-13
SHARE-44 rotate prod logs

dmetzner:
production logs are now rotated

### DIFF
--- a/config/packages/dev/monolog.yaml
+++ b/config/packages/dev/monolog.yaml
@@ -1,30 +1,69 @@
+#
+# Possible Error levels:
+#    DEBUG:     Detailed debug information.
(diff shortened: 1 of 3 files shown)


HCrane (approved):
✔️ 
Optional discussion: 
Id propose:
prod:  `max_files = 60` ✔️ (2 MOnths is okay i suppose. @bernadettespieler  thoughts?)
dev:  `max_files = 5` ❓ (actually idk but 20 log files seemed random. also i think we as devs dont need more than 5) 
test: `max_files = 5`  (who looks at these? 😛 )

dmetzner:
updated to just 5 ;)

PR #293, opened 2019-05-13
Hotfix: Jquery readded

dmetzner:
The global jquery plugin was removed with SHARE-64 but is still needed.


HCrane (approved):
:heavy_check_mark:

PR #297, opened 2019-05-18
SHARE-78 Gamjam removed

dmetzner:
removed the whole gamejam functionality from the project.


HCrane:
as discussed -> moved back to development since it is outdated

PR #298, opened 2019-05-18
SHARE-89 php7.3 support

dmetzner:
- phpUnit v8
- phpspec v5 (leanphp codecoverage report was removed from project until it supports v5)


dmetzner:
developer environments have to be updated to use php7.3 instead of 7.2

HCrane (approved):

PR #300, opened 2019-05-19
SHARE-70 XML debug/release filter

dmetzner:
The xml program upload parser now also checks for the "buildType" field.
The field is mandatory. If the field value is "debug", then the upload request sets
the programs debug_build db field to true. Else (also default for old programs) is false.

### DIFF
--- a/src/Migrations/2019/Version20190519174116.php
+++ b/src/Migrations/2019/Version20190519174116.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20190519174116 extends AbstractMigration
+{
+    public function getDescription() : string
+    {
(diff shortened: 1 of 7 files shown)


src/Migrations/2019/Version20190519174116.php
```diff
@@ -0,0 +1,43 @@
...
+        return '';
+    }
+
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE program ADD debug_build TINYINT(1) DEFAULT \'0\' NOT NULL');
+        $this->addSql('ALTER TABLE acl_classes CHANGE id id INT UNSIGNED AUTO_INCREMENT NOT NULL');
```
HCrane: Clean out this old stuff?

HCrane (changes requested):

src/Migrations/2019/Version20190519174116.php
dmetzner: done

HCrane (approved):

PR #301, opened 2019-05-20
Mastersthesis Recommender System Changes

regnisier:
In order to get more data for the online experiment, which is conducted as part of a mastersthesis, the original changes to the recommender system, which targeted personalized recommendations for users, are expanded to also effect guest users (users who are not logged in). The changes regard the recommendations on the homepage (the section Recommended Programs). The changes aim to increase the recommender systems aggregated diversity (recommend a more diverse set of items). Therefore changes to the function which recommends programs to guest users have been made (prior the recommendation list consisted of most liked programs, now a new function has been introduced in RecommenderManager.php which diversifies the former recommendation list a bit). Also the corresponding behat test has been updated.


regnisier:
Oh the newlines need to

PR #310, opened 2019-05-27
Release/v2.1.0

HCrane:
- Merge release in develop according to workflow.


HCrane:
After mergin release into develop a new PR has to be created to merge the develop into master for full release because git does apperantly not allow to make PR into two different branches :/

dmetzner (approved):

PR #317, opened 2019-06-26
SHARE-75 Prefer programs of current flavor in api calls

mfuerni:
Prefer programs tagged with the current flavor in the `api/project/`
API calls, which are used e.g. on the home page.

Adapt tests and test with two different flavors (pocketcode and luna).


dmetzner (changes requested):
When calling the api/projects/recent.json for e.g. luna the json response contains:
`"TotalProjects":20`

But when the flavor is pocketcode the json response uses a string for the number of total projects:
`"TotalProjects":"20"`

Therefore also the following test fails:
`tests/behat/features/api/get_recent_programs.feature:16`

dmetzner (approved):

PR #319, opened 2019-06-28
 SHARE-69 Filter DebugBuild programs

mfuerni:
Filter programs marked as debug build.
Show them just for user agents with build-type debug.
Adapt PHPDoc and style of several classes.
Test the behavior for debug and release app.


mfuerni:
is already based on SHARE-75, please merge #317 before to avoid merge and rebase conflicts

dmetzner (commented):
Share-75 fix needs to be included here too?

dmetzner (approved):

PR #324, opened 2019-07-23
SHARE-30 Eliminate Facebook

dmetzner:
- removed the possibility to log in via a facebook account
- removed the ability to post programs to your fb account

-- fixed a sonata update problem in the composer file. (Fixed version until deprications are fixed)


HCrane (approved):
as wanted by wolfgang and described in ticket

PR #328, opened 2019-07-24
SHARE-49 private programs are now really private

dmetzner:
- userprofile:
   statistics -> program counter includes private programs only on your own profile.
   On a different profile the counter only counts public programs.

  programs-loader -> only loads public programs for profiles except your own profile (private included)

- api:
   user program api now only returns public programs except when requesting for your own id

- program pages:
   when a program is private the program can only be found by its owner.

- also added tests


HCrane (approved):

HCrane:
@bernadettespieler needs to approve this on the testserver, talk to @dmetzner please!

PR #331, opened 2019-07-24
SHARE-109 More from <name of the user>"on the project details page

dmetzner:
- updated program details page and Programlaoder
- added some tests


HCrane (approved):
@bernadettespieler please talk to @dmetzner but from our point can be merged immediately!

PR #333, opened 2019-07-25
SHARE-108 Show categories in the sidebar

dmetzner:
Show the difference categories (featured, newest projects, etc.) in the sidebar (see media lib)


HCrane (approved):
ok from me but @bernadettespieler must check if design is as intended! (It is ok from my side)

PR #334, opened 2019-07-25
SHARE-105 Added tutorial tag

dmetzner:
NOTE: After deploying this PR the create tags command must be executed on share


HCrane (approved):
command needs to be run otherwise fine

PR #335, opened 2019-07-25
SHARE-103 project page without apk buttons for ios

dmetzner:
- removed buttons for ios project detail page
- added an twig extension that checks the user agent
- added tests


HCrane (approved):
as requested

PR #344, opened 2019-08-02
SHARE-83 Admin Flavor checks added

dmetzner:
From now on it is not possible to create/update (featured) programs/urls/mediapackages
without a valid flavor. Errormessage contains all valid flavors.


src/Admin/FeaturedProgramAdmin.php
```diff
@@ -115,7 +115,7 @@ protected function configureListFields(ListMapper $listMapper)
         'route'      => ['name' => 'show'],
         'admin_code' => 'catrowebadmin.block.programs.all',
       ])
-      ->add('flavor', 'string')
+      ->add('flavor2', 'string')
```
HCrane: wait what? :)

HCrane (changes requested):

HCrane (approved):

PR #350, opened 2019-08-13
SHARE-120 Major Symfony Upgrade

dmetzner:
- fixed all remaining deprecations to upgrade from Symfony 3.4 to 4.1
- fixed all remaining deprecations to upgrade from Symfony 4.1 to 4.2
- fixed all remaining deprecations to upgrade from Symfony 4.2 to 4.3

- **Upgraded from symfony 3.4 to 4.3**

- fixed a handful of deprecations to upgrade from Symfony 4.3 to 5

- upgraded and fixed all deprecations for most of our packages;
  finally upgraded from fr3d_ldap bundle 3 to 4!

- locked phpUnit to 8.2 -> 8.3 is buggy and fix is only in dev version

- added dmore/behat-chrome-extension (since we are using chrome headless for our tests)

- had to remove vipsoft/doctrine-data-fixtures-extension since the repo is inactive and
  not compatible with symfony 4.3. This packages reset our assets/database between every
  Behat Scenario. The same functionality (just very basic) was added to the Context files.
  (clearData() ~  @beforeScenario) This should be improved or be replaced by a different
  package in the future to reduce the runtime of the tests.

___________________

**Major change caused by the upgrade impacting commits in the future**:

Controllers are now AbstractControllers and can not access the container.
Therefore we now use (as we should have anyway) Dependency Injection.
We also use dependency injection for parameters or at least use only
the ParameterbagInterface. For more details look into the services(_test).yml
files wich are almost completly rewritten with this commit. (also with explanations :)

___________________

- fixed minor ldap login bug (gives now the correct message if the username is not found)

### DIFF
--- a/.env
+++ b/.env
@@ -38,3 +38,16 @@ SLACK_WEBHOOK=xxxx
 #DEPLOY_WEBTEST_BRANCH=
 #DEPLOY_POREVIEW=
 #DEPLOY_POREVIEW_BRANCH=
+###> symfony/framework-bundle ###
+APP_ENV=dev
+APP_SECRET=93055246cfa39f62f5be97928084989a
+#TRUSTED_PROXIES=127.0.0.1,127.0.0.2
+#TRUSTED_HOSTS='^localhost|example\.com$'
+###< symfony/framework-bundle ###
+
+###> doctrine/doctrine-bundle ###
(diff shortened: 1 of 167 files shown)


HCrane:
Merging will be put on hold until the next release to avoid having too much trouble with the current one.

PR #352, opened 2019-08-22
SHARE-122 readded & fixed codeview object image

dmetzner:
on every project page when looking at the code view there should
now be the correct object image displayed for each object.
Also added a fontawesome icon (times) when there is no image found.

### DIFF
--- a/src/Catrobat/Services/CatrobatCodeParser/ParsedObjectAsset.php
+++ b/src/Catrobat/Services/CatrobatCodeParser/ParsedObjectAsset.php
@@ -30,17 +30,8 @@ class ParsedObjectAsset
   public function __construct(\SimpleXMLElement $asset_xml_properties)
   {
     $this->asset_xml_properties = $asset_xml_properties;
-
-    if (count($asset_xml_properties->name) !== 0)
-    {
-      $this->name = $asset_xml_properties->name;
-    }
-    else
-    {
-      $this->name = $asset_xml_properties[Constants::NAME_ATTRIBUTE];
-    }
-
-    $this->file_name = $asset_xml_properties->fileName;
+    $this->name = $asset_xml_properties[Constants::NAME_ATTRIBUTE];
+    $this->file_name = rawurlencode($asset_xml_properties['fileName']);
   }
 
   /**
--- a/src/Catrobat/Services/CatrobatCodeParser/Scripts/WhenBGChangeScript.php
+++ b/src/Catrobat/Services/CatrobatCodeParser/Scripts/WhenBGChangeScript.php
@@ -25,14 +25,14 @@ class WhenBGChangeScript extends Script
 
     if (count($this->script_xml_properties->look) != 0)
     {
-      if ($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE] != null)
+      if ($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE] !== null)
       {
         $this->look_file_name = $this->script_xml_properties->look
-          ->xpath($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE])[0]->fileName;
+          ->xpath($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE])[0]['fileName'];
       }
       else
       {
-        $this->look_file_name = $this->script_xml_properties->look->fileName;
+        $this->look_file_name = $this->script_xml_properties->look['fileName'];
       }
     }
 
--- a/templates/Program/codeview.html.twig
+++ b/templates/Program/codeview.html.twig
@@ -6,7 +6,27 @@
 {% import _self as self %}
 
 {% macro dropbtn(object, path, scene_name) %}
-{#   ToDo show the correct image#}
+
+  {% if object.getLooks()|length > 0 %}
+
+    {% if scene_name is null and assetExists(path~'images/'~object.getLooks()[0].getFileName())  %}
+      <img src="{{ asset(path~'images/'~object.getLooks()[0].getFileName()) }}"
+           class="object-icon">
+    {% elseif assetExists(path~scene_name~'/images/'~object.getLooks()[0].getFileName()) %}
+      <img src="{{ asset(path~scene_name~'/images/'~object.getLooks()[0].getFileName()) }}"
+           class="object-icon">
+    {% else %}
+      <div class="object-icon">
+        <i class="no-image-found-icon fas fa-times"></i>
+      </div>
+    {% endif %}
+
+  {% else %}
+    <div class="object-icon">
+      <i class="no-image-found-icon fas fa-times"></i>
+    </div>
+  {% endif %}
+
   <div class="object-meta">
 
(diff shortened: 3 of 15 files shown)


dmetzner:
image

src/Catrobat/Services/CatrobatCodeParser/ParsedObjectAsset.php
```diff
@@ -31,16 +33,17 @@ public function __construct(\SimpleXMLElement $asset_xml_properties)
...
+      $this->file_name = $asset_xml_properties['name'];
     }
     else
     {
       $this->name = $asset_xml_properties[Constants::NAME_ATTRIBUTE];
     }
 
-    $this->file_name = $asset_xml_properties->fileName;
+    $this->file_name = $asset_xml_properties['fileName'];
+    $this->file_name = str_replace('#', '%23', $this->file_name);
```
mfuerni: Shouldn't we use `urlencode()` here instead? Or better use twig `url_encode` filter instead.

templates/Program/codeview.html.twig
```diff
@@ -6,7 +6,19 @@
 {% import _self as self %}
 
 {% macro dropbtn(object, path, scene_name) %}
-{#   ToDo show the correct image#}
+
+  {% if object.getLooks()|length > 0 %}
+    {% if scene_name is null %}
+      <img src="{{ asset('./'~path~'images/'~object.getLooks()[0].getFileName()) }}" class="object-icon">
+    {% else %}
+      <img src="{{ asset('./'~path~'/'~scene_name~'/images/'~object.getLooks()[0].getFileName()) }}" class="object-icon">
```
mfuerni: `path` ends with `/`, so use `~path~scene_name~` instead

templates/Program/codeview.html.twig
```diff
@@ -36,11 +49,27 @@
           <div class="script-text">
             <span class="script-type">{{ script.getCaption() }}</span>
           </div>
+          {% if script.getLookFileName() is defined %}
+            {% if scene_name is null %}
+              <img src="{{ asset('./'~path~'images/'~script.getLookFileName()) }}" class="brick-look script-brick-look">
+            {% else %}
+              <img src="{{ asset('./'~path~'/'~scene_name~'/images/'~script.getLookFileName()) }}"
```
mfuerni: `path` ends with `/`, so use `~path~scene_name~` instead

templates/Program/codeview.html.twig
```diff
@@ -36,11 +49,27 @@
...
         </div>
         <div class="script-bricks-wrapper">
           {% for brick in script.getBricks() %}
             <div class="brick" style="background-image: url({{ asset('images/bricks/'~brick.getImgFile()) }})">
               <div class="brick-text">{{ brick.getCaption() }}</div>
+              {% if brick.getLookFileName() is defined %}
+                {% if scene_name is null %}
+                  <img src="{{ asset('./'~path~'images/'~brick.getLookFileName()) }}" class="brick-look">
+                {% else %}
+                  <img src="{{ asset('./'~path~'/'~scene_name~'/images/'~brick.getLookFileName()) }}"
```
mfuerni: `path` ends with `/`, so use `~path~scene_name~` instead

src/Catrobat/Services/CatrobatCodeParser/ParsedObjectAsset.php
```diff
@@ -31,16 +33,17 @@ public function __construct(\SimpleXMLElement $asset_xml_properties)
   {
     $this->asset_xml_properties = $asset_xml_properties;
 
-    if (count($asset_xml_properties->name) !== 0)
+    if (count($this->file_name = $asset_xml_properties['name']) !== 0)
```
mfuerni: I don't get the idea of these two changes (if condition + body).

I think the following code is enough for this method, since name and fileName are always attributes, not nodes.
```
    $this->asset_xml_properties = $asset_xml_properties;
    $this->name = $asset_xml_properties[Constants::NAME_ATTRIBUTE];
    $this->file_name = urlencode($asset_xml_properties['fileName']);
```

mfuerni (changes requested):
The aspect ratio of the images is not kept. Maybe we should change that.
Do we need any tests for the presence of the correct image?

✓ All tests are passed

src/Catrobat/Services/CatrobatCodeParser/ParsedObjectAsset.php
dmetzner: I have tested it; it is enough. But I have used rawurlencode() which uses '%20' instead of '+' for whitespaces

dmetzner:
changed css to 'object-fit: cover' to keep the aspect ratio.
added a twig extension to check if the asset exists. If the correct images doesn't exist the object-not-found placeholder is shown.

src/Catrobat/Services/CatrobatCodeParser/ParsedObjectAsset.php
dmetzner: just had to change the outdated code.xml and unit test.

dmetzner:
- added checks for scripts, bricks and sounds, 
- fixed brick, scripts, sound getFileName
- outsourced codeview js

src/Catrobat/Services/CatrobatCodeParser/Scripts/WhenBGChangeScript.php
```diff
@@ -25,14 +25,14 @@ protected function create()
...
+      if ($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE] !== null)
       {
         $this->look_file_name = $this->script_xml_properties->look
-          ->xpath($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE])[0]->fileName;
+          ->xpath($this->script_xml_properties->look[Constants::REFERENCE_ATTRIBUTE])[0]['fileName'];
       }
       else
       {
-        $this->look_file_name = $this->script_xml_properties->look->fileName;
+        $this->look_file_name = $this->brick_xml_properties->look['fileName'];
```
mfuerni: `brick_xml_properties` does not exist in class `Script`. Must be changed to `script_xml_properties`.

templates/Program/codeview.html.twig
```diff
@@ -37,19 +58,37 @@
...
+                   class="brick-look script-brick-look">
+            {% endif %}
+          {% endif %}
         </div>
         <div class="script-bricks-wrapper">
           {% for brick in script.getBricks() %}
             <div class="brick" style="background-image: url({{ asset('images/bricks/'~brick.getImgFile()) }})">
               <div class="brick-text">{{ brick.getCaption() }}</div>
+              {% if brick.getLookFileName() is defined %}
+                <img src="{{ asset('./'~path~scene_name~'/images/'~brick.getLookFileName()) }}"
```
mfuerni: I think this line (75) should be removed.

mfuerni (changes requested):
Just two minor fixes, then I can approve it.

mfuerni (approved):

PR #353, opened 2019-08-22
SHARE-117 added change project screenshot/thumbnail functionality

dmetzner:
A user can now change the screenshot/thumbnail of his projects similar to
the way he can change his profile avatar.
    
- created imageUtils.php and Imageupload.js to prevent duplicated code.
- the uploaded image gets checked and resized to the correct sizes (same like app upload)
- img gets refreshed instant on the html page, no need to refresh the page
- same error messaged like @myprofile avatar + 1 success message


dmetzner:
merge after SHARE-120

HCrane (approved):

PR #355, opened 2019-08-25
SHARE-118 Allow Webview Login using Cookies

mfuerni:
Allow to log in automatically inside the app by using a token.
Hide logout button if logged in using cookies.
Show advice to login again using the app if authentication fails.

To log in, set cookie `CATRO_LOGIN_USER` to the **URL-encoded username**
and cookie `CATRO_LOGIN_TOKEN` to the token returned by API login.

Fix static (not translated; exception not caught) error message
in `ApiKeyAuthenticator::authenticateToken`.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -79,6 +79,11 @@ services:
     public: true
   App\Catrobat\Security\ApiKeyAuthenticator: "@apikey_authenticator"
 
+  webview_authenticator:
+    class: App\Catrobat\Security\WebviewAuthenticator
+    public: true
+  App\Catrobat\Security\WebviewAuthenticator: "@webview_authenticator"
+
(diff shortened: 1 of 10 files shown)


config/services.yaml
```diff
@@ -79,6 +79,13 @@ services:
     public: true
   App\Catrobat\Security\ApiKeyAuthenticator: "@apikey_authenticator"
 
+  webview_authenticator:
+    class: App\Catrobat\Security\WebviewAuthenticator
+    arguments:
```
dmetzner: arguments section is not needed; Dependency injection of TranslatorInterface in the constructor of webview_authenticator is already enough.

dmetzner (approved):

PR #363, opened 2019-09-04
SHARE-143 Search Behat test fails (cons.searches ...)

matt-sp:
Reason of failed test: Search button did not work due to Javascript error.

Fixed Javascript null error caused due to a wrong reference to the search input field in analytics.js


dmetzner (approved):
Requirements ✔️
Tests ✔️

PR #376, opened 2019-11-20
Share 114 implement docker

fladens:
Implementation of Docker

### DIFF
--- a/docker/config/parameters.yml.dist
+++ b/docker/config/parameters.yml.dist
@@ -1,0 +1,44 @@
+parameters:
+    database_driver:   pdo_mysql
+    database_host:     db.catroweb.dev
+    database_port:     3306
+    database_name:     symfony
+    database_user:     root
+    database_password: root
+    database_server_v: 10.3.11
+
+    locale:            en
+    secret:            ThisTokenIsNotSoSecretChangeIt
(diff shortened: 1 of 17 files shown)


docker/config/parameters.yml.dist
```diff
@@ -0,0 +1,44 @@
+parameters:
+    database_driver:   pdo_mysql
+    database_host:     db.catroweb.dev
+    database_port:     3306
+    database_name:     symfony
+    database_user:     root
+    database_password: root
+    database_server_v: 10.3.11
```
HCrane, line 8: it should be `database_server_v: mariadb-10.3.11`

docker/php/LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
```
HCrane: Out of curiosity is this license needed?

HCrane:
Something i noticed when trying to do composer updates -> the `composer.json` and `composer.lock` do not get changed so a real update is not done only inside the containers...this should change.

dmetzner (approved):

HCrane:
Rebased to latest develop, all test run fine natively so we are good to go for merge!

HCrane (approved):

PR #380, opened 2019-11-23
SHARE-167 Behat Tests HTTPS and APP_ENV

mfuerni:
API Behat tests using iInvokeTheRequest() were using https always.
Use protocol based on $this->server_parameters.
Fix tests that assume https.

Behat tests using Symfony2Extensions accessed Symfony directly,
using APP_ENV=dev. Therefore, loading .env.dev instead of .env.test
Fix this by using a separate bootstrap_test.php file.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -25,7 +25,7 @@ default:
 
         Behat\Symfony2Extension:
             kernel:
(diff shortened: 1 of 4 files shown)


HCrane (approved):
Ran the tests twice on my machine and works accordingly!

PR #381, opened 2019-11-27
Intro

IsolatedSys:
Introduction Task:
Be aware of test

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -25,7 +25,7 @@ default:
 
         Behat\Symfony2Extension:
             kernel:
-                bootstrap: config/bootstrap.php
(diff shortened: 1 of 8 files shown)


dmetzner (commented):
The introduction task looks fine to me and you can start working on real tickets :)

Still some sidenotes:

- When creating a PR make sure to rebase to the develop branch. You should only have 1 commit containing only your changes. You might try this with this PR before closing it.

- For future tickets write the JS code in .js files and use grunt

-  For the tests:
 try to avoid stuff like this: "Then I wait 2000 milliseconds"
at least keep the number as low as possible, else the tests will take for ever

PR #386, opened 2019-12-18
SHARE-157 Example projects for Embroidery and Arduino

FelixAuf:
- Admins can add example programs via the admin panel, these programs will be visible in the chosen flavor (either Embroidery or Arduino), similarly to the featured programs.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9495fcaf8ab3d2935aa9fd2a2ae73c58",
+    "content-hash": "1f551fa954643357cecb9f92830a8304",
     "packages": [
         {
             "name": "catrobat/capi",
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -19,7 +19,7 @@ security:
 
         api_checktoken_v2:
             provider: chain_provider
-            pattern: ^.*/api/checkToken
+            pattern: ^/api/checkToken
             stateless: true
             guard:
                 authenticators:
@@ -27,7 +27,7 @@ security:
 
         api_upload_v2:
             provider: chain_provider
-            pattern: ^.*/api/projects/upload
+            pattern: ^/api/projects/upload
             stateless: true
             guard:
                 authenticators:
@@ -35,7 +35,7 @@ security:
 
         api_logout_v2:
             provider: chain_provider
-            pattern: ^.*/api/logout
+            pattern: ^/api/logout
             stateless: true
             guard:
                 authenticators:
@@ -43,7 +43,7 @@ security:
 
         api_private_user_projects_v2:
             provider: chain_provider
-            pattern: ^.*/api/projects/user/*
+            pattern: ^/api/projects/user/*
             stateless: true
             guard:
                 authenticators:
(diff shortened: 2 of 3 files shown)


templates/Index/index.html.twig
```diff
@@ -29,6 +29,27 @@
 
   </div>
 
+  {%  if theme() == 'arduino' or theme() == 'embroidery' %}
+    <div id="examplePrograms">
+      <h1>{{ "programs.example"|trans({}, "catroweb") }}</h1>
+
+      {% if example|length > 0 %}
+        {{ include('Index/example_programs_carousel.html.twig',
```
dmetzner: use featured_programs_carousel.html.twig ( might rename it since it now has multiple usages)

templates/Index/example_programs_carousel.html.twig
```diff
@@ -0,0 +1,43 @@
+<div id="{{ carousel.id }}" class="carousel slide {{ carousel.classes }} center" data-ride="carousel">
```
dmetzner: get rid of this file, it just a copy of the features programs slider

src/Entity/ExampleProgram.php
```diff
@@ -0,0 +1,223 @@
...
+
+  /**
+   * @ORM\ManyToOne(targetEntity="Program", fetch="EAGER")
+   **/
+  private $program;
+
+  /**
+   * @ORM\Column(type="string", nullable=true)
+   */
+  protected $url;
```
dmetzner: for what is this url member used? I found no field to set it in the admin interface.

src/Catrobat/Controller/Web/DefaultController.php
```diff
@@ -75,6 +84,15 @@ public function indexAction(Request $request, FeaturedImageRepository $image_rep
       $featured_items = $repository->getFeaturedItems($flavor, 10, 0);
     }
 
+    if ($flavor === 'arduino')
+    {
+      $example_items = $example_repository->getExampleItems('arduino', 10, 0);
+    }
+    else
```
dmetzner: only load embroidery projects when the embroidery flavor is used,

might be better to just load examples for the current flavor, if flavors don't have examples just don't show this category. ...->getExampleItems($flavor, 10, 0)

src/Catrobat/Controller/Api/ExampleController.php
```diff
@@ -0,0 +1,120 @@
...
+   * @throws NonUniqueResultException
+   */
+  public function getExampleProgramsAction(Request $request, ExampleImageRepository $image_repository,
+                                            ExampleRepository $repository)
+  {
+    return $this->getExamplePrograms($request, false, $image_repository, $repository);
+  }
+
+
+  /**
```
dmetzner: it would be better to just use one route and retrieve ios from the request.
like it will be done for featured projects

image

@HCrane you might have to add this API call to your Swagger API.
Or should this be changed later with the API rework? @HCrane

dmetzner (changes requested):
Tests passed.
Overall this PR looks quite nice. Just a few minor things:

- add the example dir to the project + a .gitignore else the project is corrupted for everyone else
image

- Remove the Code Duplication for the carousel.

- when the examples category is empty, also hide the title (like it is done for almost all other index categories; might also fix this for the featured projects)

- in the admin interface, it would be nice to see the image constraints before trying to upload an image with an invalid size. But this has very low priority, only do this if it can be done easily. 

- might implement the Api in a way as it was discussed for featured projects at the API meeting?@HCrane

PR #389, opened 2020-01-03
SHARE-169 Comments shown on remixed projects

matt-sp:
For a program, random comments not belonging to it, were shown.

The reason was a column type mismatch (int <> guid) in the database
model / ORM. GUID was converted to INT resulting in wrong and random
relations between comments and programs.

- Fixed column types
- Fixed entity's setter and getter and corresponding code parts
calling them
- Created doctrine migration
- Extended behat scenario to test the changes

### DIFF
--- a/tests/behat/features/web/comments.feature
+++ b/tests/behat/features/web/comments.feature
@@ -73,6 +73,7 @@ Feature: As a visitor I want to write, see and report comments.
   Scenario: I should be able to see existing comments
     Given I am on "/app/project/1"
     Then I should see "c1"
+    Then I should not see "c2"
 
   Scenario: I should not see any comments when there are none
     Given I am on "/app/project/3"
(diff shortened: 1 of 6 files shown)


tests/behat/features/web/comments.feature
```diff
@@ -189,6 +190,8 @@ Feature: As a visitor I want to write, see and report comments.
     And I wait 250 milliseconds
     When I click "#comment-post-button"
     And I wait for a second
+    And I wait for a second
+    And I wait for a second
```
HCrane: Please remove those 2 lines!

src/Migrations/2020/Version20200103175935.php
```diff
@@ -0,0 +1,77 @@
+<?php
```
HCrane: As discussed plese regenerate the migration!

HCrane (changes requested):

HCrane (approved):

PR #422, opened 2020-02-03
Deleting a user should also delete associated notifications

matt-sp:
Currently when deleting a user with (like) notifications an error will
be displayed

- Created Doctrine annotations to facilitate cascade remove.

### DIFF
--- a/src/Entity/CatroNotification.php
+++ b/src/Entity/CatroNotification.php
@@ -39,7 +39,7 @@ class CatroNotification
   /**
    * @var User
    *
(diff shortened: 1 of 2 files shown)


dmetzner (changes requested):
Based on your commit message. Are you sure you pushed all the changes? I can't see any changed behat tests / delete action.

Also it still doesn't work to delete a user who has a (follow)notification.

image

how to replicate:
follow userA with userB
log in with userA and got to  "my profile"
delete my account (userA)

PR #429, opened 2020-02-06
SHARE-119 Remix Notifications

leno12:
-wrote more testcases and fixed some bugs


dmetzner (approved):
The remaining RemixBug will be fixed in the ticket: 
https://jira.catrob.at/browse/SHARE-203

dmetzner:
SHARE-203 is now also fixed in this PR: 
including uuid extraction for remix projects (including a new test)
spec tests to work with strings as id
old vs new project comparison is now based on creation date instead of id

PR #430, opened 2020-02-10
Introduction Task

JackShahhoud:


src/Catrobat/StatusCode.php
```diff
@@ -41,6 +41,8 @@ class StatusCode
   const PASSWORD_INVALID = 532;
   const CREDITS_TO_LONG = 833;
 
+  const ALREADY_STOLE = 510;
```
dmetzner, line 44: try not to override codes that are already defined https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/510

dmetzner (commented):
Well done introduction task. Have fun working on the real issues! :)

You can close this PR now

PR #431, opened 2020-02-11
SHARE-157 Example Projects

FelixAuf:
Admins can add example programs via the admin panel, these programs will be visible in the chosen flavor (either Embroidery or Arduino), similarly to the featured programs. Headers for "Featured" and "Examples" are set invisible if there are no such projects. Also added help text for image constraints in Sonata.


dmetzner (changes requested):
As discussed.

dmetzner (changes requested):
The git problems must be fixed

tests/behat/features/admin/example_projects.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  Scenario: List all programs
+    Given I log in as "Adminius" with the password "123456"
+    And I am on "/admin/example_program/list"
+    And I wait for the page to be loaded
+    Then I should see the example table:
+      | Id | Name      | Flavor     | Example |
+      | 1  | program 1 | pocketcode | yes     |
+      | 2  | program 2 | pocketcode | no      |
+      | 3  | program 3 | pocketcode | no      |
```
dmetzner: add a test:
- set a project to an example project 
- reset an example project to a normal project
- change the flavor of example project

dmetzner (changes requested):
1) Please add more tests for the admin interface
2) The new API also has the possibility to ask for all example projects.
    Add tests similar to `tests/behat/features/api/projects/get_most_downloaded_projects.feature` for the example projects.

3) There seems to a bug concerning the flavor change:

luna has no newest project section. But when I switch an example project from pocketcode to luna - luna suddenly has a newest project + recommended section on the homepage in addition to the example section.

before:
image
 after:
image

dmetzner (changes requested):
Changes requested from po review:

- only list example projects
- add a mask for new example project creation
 -- option to set a flavor (this allow to use the same project for different flavors as example project)
 -- option to set a thumbnail directly (every flavor might use a different thumbnail for the same project)
should be similar to Featured projects...
image

src/Entity/Program.php
```diff
@@ -1022,4 +1025,9 @@ public function setRemixNotificationMentionsAsParent(Collection $remix_notificat
   {
     $this->remix_notification_mentions_as_parent = $remix_notification_mentions_as_parent;
   }
+
+  public function isExample(): bool
+  {
+    return false;
```
dmetzner, line 1031: always returning false? Was this intended?

templates/Index/index.html.twig
```diff
@@ -53,6 +52,11 @@
       <div class="programs"></div>
     </div>
 
+    <div id="example" style="display:none">
+      <h1>{{ "programs.example"|trans({}, "catroweb") }}</h1>
```
dmetzner:
```
should be a <h2>

there must be only one h1 per page
```

package-lock.json
```diff
@@ -7,7 +7,7 @@
     "@babel/code-frame": {
```
dmetzner: We should really try to update npm only in explicit PRs. Else there will always be merge conflicts.

src/Catrobat/Controller/Api/ListProgramsController.php
```diff
@@ -45,6 +45,20 @@ public function listProgramIdsAction(Request $request): ProgramListResponse
     return $this->listSortedPrograms($request, 'recent', false);
   }
 
+  /**
```
dmetzner, line 48: the new Projects API is already finished - no need to add something to the old one.
this call can be removed.

src/Catrobat/Controller/Api/ListProgramsController.php
```diff
@@ -191,6 +205,10 @@ private function listSortedPrograms(Request $request, string $sortBy, bool $deta
         $flavor, $limit, $offset, $max_version
       );
     }
+    elseif ('example' === $sortBy)
```
dmetzner, line 208: also not needed

src/Catrobat/Forms/ExampleImageConstraint.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace App\Catrobat\Forms;
+
+use Symfony\Component\Validator\Constraint;
+
+class ExampleImageConstraint extends Constraint
+{
+  public int $required_width = 512;
```
dmetzner: I think this constraint can be dropped.
Project thumbnails are "optimized" during the project upload. (take a look into the ScreenshotRepository.php class)

If possible reuse the same methods. 

@SasaDermanovic What is your opinion about it? Since you are reworking the image optimizations.

src/Catrobat/Listeners/View/ProgramListSerializer.php
```diff
@@ -65,8 +80,16 @@ public function onKernelView(ViewEvent $event): void
           $new_program['Uploaded'] = $program->getUploadedAt()->getTimestamp();
           $new_program['UploadedString'] = $this->time_formatter->getElapsedTime($program->getUploadedAt()
             ->getTimestamp());
-          $new_program['ScreenshotBig'] = $this->screenshot_repository->getScreenshotWebPath($program->getId());
-          $new_program['ScreenshotSmall'] = $this->screenshot_repository->getThumbnailWebPath($program->getId());
+          if ($new_program['Example'])
+          {
```
dmetzner, line 81: that's old API and not really needed. Also, most of the logic would not have been needed if the projects thumbnails were initialized correctly

src/Catrobat/Services/ExampleImageRepository.php
```diff
@@ -0,0 +1,68 @@
...
+    $dir = preg_replace('#([^\/]+)$#', '$1/', $dir);
+    $path = preg_replace('#([^\/]+)$#', '$1/', $path);
+
+    if (!is_dir($dir))
+    {
+      throw new InvalidStorageDirectoryException($dir.' is not a valid directory');
+    }
+
+    $this->dir = $dir;
+    $this->path = $path;
```
dmetzner:
```
phpcpd 5.0.2 by Sebastian Bergmann.

Found 1 clones with 28 duplicated lines in 2 files:

  - /github/workspace/src/Catrobat/Services/ExampleImageRepository.php:22-50 (28 lines)
    /github/workspace/src/Catrobat/Services/FeaturedImageRepository.php:22-50

0.06% duplicated lines out of 43502 total lines of code.
Average size of duplication is 28 lines, largest clone has 28 of lines

```
https://github.com/Catrobat/Catroweb-Symfony/pull/431/checks?check_run_id=642604864

src/Catrobat/Services/ExampleImageRepository.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace App\Catrobat\Services;
+
+use App\Catrobat\Exceptions\InvalidStorageDirectoryException;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+use Symfony\Component\HttpFoundation\File\File;
+use Symfony\Component\HttpFoundation\UrlHelper;
+
+class ExampleImageRepository
```
dmetzner:
```
 ------ ----------------------------------------------------------------------- 
  Line   src/Catrobat/Services/ExampleImageRepository.php                       
 ------ ----------------------------------------------------------------------- 
  37     Parameter #1 $id of method                                             
         App\Catrobat\Services\ExampleImageRepository::generateFileNameFromId(  
         ) expects string, int given.                                           
  42     Parameter #1 $id of method                                             
         App\Catrobat\Services\ExampleImageRepository::generateFileNameFromId(  
         ) expects string, int given.                                           
  56     Parameter #1 $id of method                                             
         App\Catrobat\Services\ExampleImageRepository::generateFileNameFromId(  
         ) expects string, int given.                                           
 ------ ----------------------------------------------------------------------- 
```

src/Catrobat/Listeners/View/ProgramListSerializer.php
```diff
@@ -51,6 +55,17 @@ public function onKernelView(ViewEvent $event): void
       foreach ($programs as $program)
       {
         $new_program = [];
+        if ($program->isExample())
+        {
+          $new_program['ExampleId'] = $program->getId();
+          $new_program['Extension'] = $program->getImageType();
```
dmetzner, line 62:
```
 ------ ----------------------------------------------------------------- 
  Line   src/Catrobat/Listeners/View/ProgramListSerializer.php            
 ------ ----------------------------------------------------------------- 
  61     Call to an undefined method App\Entity\Program::getImageType().  
  63     Call to an undefined method App\Entity\Program::getProgram(). 
```

dmetzner (changes requested):
I have highlighted some minor stuff.

All in all, it already works quite nicely. Also great that it is already integrated and tested in the new API.

- the old api stuff should be removed - then also the old api tests should pass again.
- ...

src/Catrobat/Forms/ExampleImageConstraint.php
SasaDermanovic: Yes, that would be my verdict too, since the constraint is irrelevant and therefore can be dropped.

dmetzner (approved):

PR #435, opened 2020-02-12
 SHARE-159 Show number of rejected programs/comments per user

wolfianer:
Implemented "Reported Users" in the admin panel.

### Short Description
Implemented "Reported Users" in the admin panel.

As written in discord some behat tests fail for the admin panel, because I added the working Assert::assert() function instead of the not working assert() function. 

[FailedTests.txt](https://github.com/Catrobat/Catroweb-Symfony/files/4191967/FailedTests.txt)

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -4,6 +4,7 @@ namespace App\Entity;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\Criteria;
 use Doctrine\DBAL\Types\GuidType;
 use FR3D\LdapBundle\Model\LdapUserInterface;
 use Sonata\UserBundle\Entity\BaseUser as BaseUser;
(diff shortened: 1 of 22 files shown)


src/Entity/User.php
```diff
@@ -452,4 +464,46 @@ public function isFollowing($user)
   {
     return $this->following->contains($user);
   }
+
+  /**
+   * @return mixed
+   */
+  public function  getProgramInappropriateReports()
```
dmetzner: 2 whitespaces

src/Entity/User.php
```diff
@@ -452,4 +464,46 @@ public function isFollowing($user)
...
+    {
+      $count +=$program->getReportsCount();
+    }
+    return $count;
+  }
+
+  /**
+   * @return mixed
+   */
+  public function  getComments()
```
dmetzner: again

dmetzner (changes requested):
Nice PR, works as it should, only two minor things:

* fix the mapping errors (can be found in the symfony profiler) 

image

* might re-format User.php. There are two necessary whitespaces.

Tests Are still running. I will update this comment when they are done.

dmetzner:
perfect, now just squash it, then we can merge it

dmetzner (approved):

dmetzner (approved):

PR #510, opened 2020-04-21
SHARE-197 API Search

JackShahhoud:
- Implement API Search
- Update/Fix random projects
- Update API response body
- Update API functions
- Refactor API tests
---

- [x] Include the name and id of the Jira ticket in the PR’s title

### DIFF
--- a/src/Api/MediaLibraryApi.php
+++ b/src/Api/MediaLibraryApi.php
@@ -9,6 +9,7 @@ use App\Entity\MediaPackageFile;
 use Doctrine\ORM\EntityManagerInterface;
 use OpenAPI\Server\Api\MediaLibraryApiInterface;
 use OpenAPI\Server\Model\MediaFile;
+use OpenAPI\Server\Model\MediaFiles;
 use OpenAPI\Server\Model\Package;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\Routing\Generator\UrlGenerator;
@@ -72,6 +73,14 @@ class MediaLibraryApi implements MediaLibraryApiInterface
    */
   public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 20, ?int $offset = 0, &$responseCode, array &$responseHeaders)
   {
+    if (null === $limit)
+    {
+      $limit = 20;
+    }
+    if (null === $offset)
+    {
+      $offset = 0;
+    }
     $media_package = $this->entity_manager->getRepository(MediaPackage::class)
       ->findOneBy(['nameUrl' => $package_name])
     ;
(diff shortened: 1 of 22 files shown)


wolfianer:
Looks good, after the first glance.
Locally some tests fail after rebasing, so pls rebase.

wolfianer (approved):

tests/behat/features/api/mediapackage/mediafile_search_get.feature
```diff
@@ -147,7 +147,7 @@ Feature: Get data from the media library in json format
     And I have a parameter "query_string" with value "Cat"
     And I have a parameter "flavor" with value "pocketcode"
     And I request "GET" "/api/media/file/search"
-    Then the response status code should be "200"
+    Then The response status code should be "200"
```
dmetzner: "the" would be the best practice not "The"

The sentence starts with the keyword (Then, When, And, ...) 
so the first word of the step definition is not the first word of the sentence.

Therefore it should be in lower case, with exceptions as we know them from the english grammar.

https://docs.behat.org/en/latest/user_guide/gherkin.html

tests/behat/context/ApiContext.php
```diff
@@ -1635,6 +1650,256 @@ public function iShouldGetTheStoredJsonObject(string $name): void
...
+      }
+      $projects = implode(',', $values);
+      Assert::assertEquals(count($expected_programs), 0, 'Expected projects have ['.$projects.'] unmatched');
+    }
+  }
+
+  /**
+   * @Then /^The response should contain projects in the following order:$/
+   */
+  public function responseShouldContainProjectsInTheFollowingOrder(TableNode $table): void
```
dmetzner, line 1712: All those new step definitions are quite nice since they highly reduce the amount of necessary test code.
But, and that is a big one. Now we are not testing with a way lower precision.  So if we are going to keep them there is still much work to do:

Especially in the API, we must not check that project names are returned in the correct order. Every field must be checked for the correct content. Also, it must be checked that all fields are returned and nothing more. 
Therefore I'm not even sure if it would be enough to compare objects after the json_decode.

In my opinion, we should not lower the testing precision concerning the API. Minor changes we are not tracking could result in broken code by other teams using our API. To immediately see changed behavior, I see no other possibility than to compare the whole JSON response with at least an expected JSON response based on the projects (It can still be generated). 

The same applies to all these step definitions.

src/Api/MediaLibraryApi.php
```diff
@@ -82,44 +91,65 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+            'download_url' => $this->url_generator->generate(
+              'download_media',
+              ['id' => $media_package_file->getId()],
+              UrlGenerator::ABSOLUTE_URL),
+          ];
+
+          $json_response_array[] = new MediaFile($mediaFile);
         }
       }
     }
```
dmetzner, line 149: why change this API? it should be already in a working state - is there a bug?

dmetzner (changes requested):
In the current state, I'm against merging it. @HCrane your opinion about the tests would be highly appreciated.

tests/behat/context/ApiContext.php
```diff
@@ -1635,6 +1650,271 @@ public function iShouldGetTheStoredJsonObject(string $name): void
...
+
+  /**
+   * @Then /^the response should have the featured projects model structure$/
+   */
+  public function responseShouldHaveFeaturedProjectsModelStructure(): void
+  {
+    $response = $this->getKernelBrowser()->getResponse();
+    $responseArray = json_decode($response->getContent(), true);
+
+    foreach ($this->programs_structure as $key)
```
dmetzner, line 1832: count check is missing:
projects total_results, ...

tests/behat/context/ApiContext.php
```diff
@@ -1635,6 +1650,271 @@ public function iShouldGetTheStoredJsonObject(string $name): void
...
+    {
+      Assert::assertArrayHasKey($key, $responseArray, 'Response should contain '.$key);
+    }
+
+    $returned_programs = $responseArray['projects'];
+
+    foreach ($returned_programs as $program)
```
dmetzner, line 1839: might wanna rename program in this scenario to something like
featured_program_response. Since this object contains way less field than a Program $program object. 
Else it can be a little confusing at first.

also in the checkFeaturedProjectFieldsValue method and so on

tests/behat/context/ApiContext.php
```diff
@@ -3069,6 +3351,159 @@ public function iUploadAStandardCatrobatProgram(): void
...
+        Assert::assertIsString($author);
+      },
+      'featured_image' => function ($featured_image)
+      {
+        Assert::assertIsString($featured_image);
+        Assert::assertRegExp('/http:\/\/localhost\/resources_test\/featured\/featured_[0-9]+\.jpg/',
+          $featured_image);
+      },
+    ];
+    call_user_func($fields[$key], $program[$key]);
```
dmetzner, line 3553: Have you tried out what happens if the method is called with a key that does not exist? I'm interested.

tests/behat/context/ApiContext.php
```diff
@@ -1635,6 +1650,271 @@ public function iShouldGetTheStoredJsonObject(string $name): void
...
+
+    $responseArray = json_decode($response->getContent(), true);
+    $returned_programs = $responseArray['projects'];
+    $expected_programs = $table->getHash();
+
+    Assert::assertEquals(count($returned_programs), count($expected_programs), 'Number of returned programs should be '.count($returned_programs));
+
+    for ($i = 0; $i < count($returned_programs); ++$i)
+    {
+      Assert::assertEquals($returned_programs[$i]['name'], $expected_programs[$i]['Name']);
```
dmetzner: we still only compare the name -
we should really also check the values of all other fields.

My suggestion: 

- don't define the name in the feature file - use the unique id
- get the project from the database
- for every structure key check the value with the returned one
- some fields like the urls might need a little extra logic

dmetzner (changes requested):

src/Entity/ProgramManager.php
```diff
@@ -823,6 +843,26 @@ public function getProjects(string $project_type, string $max_version = '0',
...
+    switch ($project_type){
+      case 'recent':
+        return $this->getRecentProgramsCount($flavor, $max_version);
+      case 'random':
+        return $this->getRandomProgramsCount($flavor, $max_version);
+      case 'most_viewed':
+        return $this->getMostViewedProgramsCount($flavor, $max_version);
+      case 'most_downloaded':
+        return $this->getMostDownloadedProgramsCount($flavor, $max_version);
+      case 'example':
```
dmetzner, line 861: Example projects are finally merged -would be perfect if you could add them

dmetzner (changes requested):
Looks fine to me. Just fix the 1 test that is failing.

dmetzner (approved):

PR #517, opened 2020-04-25
SHARE-238 Refactor Adminarea Projects | All projects

leno12:
- Programs are now sorted by upload date not by id
- Removed mosaic view
- Removed Downloads Filter and added option to filter by upload date (DateTime Picker)
- Wrote new test cases for all projects section in admin area
- Disabled sorting options for name, visible, approved
- Removed description field since this just needs to be basic overview of the project (more details can be seen by clicking on the show button)
- Added Filter by Flavor Option
- Added option to select flavor when changing flavor so that user cannot enter wrong flavor
- Fixed size of the thumbnail images
- Removed Edit Button (we can still edit the project by clicking on the project name)

---

### Additional Description


hebesempso (changes requested):
The changes and tests you made look good. 

But I have found some things that should be discussed:
- Does the id filter still make any sense? Because there are no id's displayed in the table
- Does changing the program name, description, user who made the program, #downloads, #views and flavour make any sense or is it useful in any way? I don't know if you've talked with the seniors about the behaviour of these fields and how they should work @dmetzner @HCrane (also if they are not useful, the edit button wouldn't be needed anymore, because one can change visible and accepted in the table anyway)
- If you change the flavor of a program to one that doesn't exist, the edit page will crash when editing something else (for example the name of the project). Maybe a choice of all available flavours would be useful (like if you want to change  the user, you can pick between all exisiting ones)
- If an admin wants to change the visibility to not visible, it seems that this would be an action like reporting the program, so maybe instead of changing the visibility, it should report the program (and should automatically set report state to accepted and visibility to not visible). Again this is a thing, that the seniors should decide about

dmetzner:
1) id filter - I would say yes. One might need to find a specific project. IDs are unique - A name isn't.
2) I can't think of anything useful. Might if a user has a new account and asks for the transfer of his old projects? At least views and downloads should not be changeable. Else we might have some admins boosting their stats ;) . Description and name too - I see no reason to change them.  @hcrane you know more about this topic?
3) a dropdown or something would be perfect yes
4) I am not sure about this - A report implies invisible yes, but invisible does not imply reported. Admins might just wanna set (featured) projects invisible until they are ready for the public. On the other hand, such projects could just be set to private.@hcrane your opinion?

dmetzner (changes requested):
- remove sort by name, description flavor
  -- sorting by views/downloads is useful - keep them

- remove description field - this section should just be an overview - more details are anyway behind the show action

- the flavor should only accept correct flavor values.

- instead of the project name it would be nice to have it as we have it at the reported section:
image. This way, the id and name are shown combined

- is it possible to have same sized thumbnails? The no-screenshot ones are very large. 
  If it is not easily possible might remove the NoScreenshot image completely. or maybe it is possible to set a max size?
image

--

Show button:
  - leave it as it is


Edit button:
- completely remove it

Filter:
- change Filter text from "User Username" to just "Username"
- add a flavor filter
- keep the id filter

and pls rebase


The Edit Action itself - I would say we handle it in its own ticket.

package.json
```diff
@@ -33,24 +33,24 @@
...
   "dependencies": {
-    "ajv": "^6.12.0",
+    "ajv": "^6.12.2",
     "animate.css": "^3.7.2",
     "animatedmodal": "^1.0.0",
     "bootstrap": "^4.4.1",
     "clipboard": "^2.0.6",
     "grunt-contrib-uglify-es": "github:gruntjs/grunt-contrib-uglify#harmony",
-    "jquery": "^3.4.1",
+    "jquery": "^3.5.0",
```
dmetzner: This will result in failing checks as soon as you have rebased.
https://github.com/Catrobat/Catroweb-Symfony/pull/527

dmetzner (approved):
Just rebase again :)

PR #573, opened 2020-05-29
SHARE-300 Implement new OAUTH Provider

leno12:
- Installed HWIOAUTH Bundle
- Configured all yml files needed for HWIOAUTH
- Implemented FOSUBUserProvider class to handle login using Oauth
- New migration version with fields for google id and access token

---

### Additional Description


dmetzner:
- we must set the configs correct on the auto-deploy server + enable the origin in the request in Google developer console to really check the feature

- Mark the old social members of a user (g+ tokens, ..) as deprecated.

- The API must also be renewed. Take a look at the security controller / oauth service.
Try to outsource the API interface to CAPI and make sure to explicitly define everything that is needed by the other teams (android) to make the oauth api work. (Could be part of a different ticket, if the current is api is not broken by this changes)

- try not to implicitly upgrade dependencies in package.json/composer.json if not needed. Else there will be many merge conflicts. That's basically what we have the dependabot now for.

translations/catroweb.en.yml
```diff
@@ -66,6 +67,11 @@ notificationsClearError: "Unable to mark notifications as read!"
 notificationsDeleteError: "Unable to delete notifications!"
 pleaseWait: "Please wait..."
 projects: "Projects"
+oauthPopUpTitle: "Oauth Sign In Information"
```
dmetzner: Please use correct english grammar. (uppercase/lowercase)

**OA**uth **s**ign **i**n **i**nformation

translations/catroweb.en.yml
```diff
@@ -66,6 +67,11 @@ notificationsClearError: "Unable to mark notifications as read!"
...
+oauthInfo: 'You have just used Oauth system to sign in. We created a random username for you. If you want
```
dmetzner: Do user understand oauth? or should we think of a term even kids understand?
any ideas?

src/Catrobat/Security/FOSUBUserProvider.php
```diff
@@ -0,0 +1,121 @@
...
+
+    for ($i = 0; $i < $length; ++$i)
+    {
+      try
+      {
+        $pass .= $chars[random_int(0, $max)];
+      }
+      catch (Exception $e)
+      {
+      }
```
dmetzner: this catch might be dangerous.
Maybe it is better if we let the application crash in this scenario - or use a different approach for the password in this case?

In the worst case we might get users without passwords. The exception should not happen anyway. So maybe just remove the try/catch

src/Catrobat/Security/FOSUBUserProvider.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    return $pass;
+  }
+
+  private function createRandomUsername(): string
+  {
+    $user = 'user';
+    $user_number = 1;
```
dmetzner: similar to the reason we use uuids.

The management does not like this fixed incremented ids for various reasons.

On the other hand this process will be very slow once we have thousands of OAuth users.

It may does not look that pretty but we should try to use some kind of randomness in this naming scheme.

```
id = rand(0, 99999999999999)
hash_id = sha(id) 
```

something in this direction maybe? Still we should try to keep the usernames quite small, else most design aspects on the webpage look awful as soon as we show a username somewhere.

What's your view on this?  I will definetly ask UX for their opinion.

src/Catrobat/Controller/Web/DefaultController.php
```diff
@@ -160,4 +162,29 @@ public function makeNonRecommendedProgramClickStatisticAction(Request $request):
...
+   */
+  public function checkOauthFirstLogin(): Response
+  {
+    /** @var User|null $user */
+    $user = $this->getUser();
+    $user_first_login = false;
+    $user_id = null;
+    if (null !== $user)
+    {
+      if (true == $user->isOauthUser() && !$user->isOauthPasswordCreated())
```
dmetzner: no need for the nested if, one is enough

src/Catrobat/Controller/Web/DefaultController.php
```diff
@@ -160,4 +162,29 @@ public function makeNonRecommendedProgramClickStatisticAction(Request $request):
...
+      {
+        $user_first_login = true;
+        $user_id = $user->getId();
+      }
+    }
+
+    return JsonResponse::create([
+      'statusCode' => Response::HTTP_OK,
```
dmetzner: a HTTP response already hast a response code. No need to artificially add this 'statusCode' entry

dmetzner: I think it is the second param. of a JsonResponse

dmetzner (changes requested):
I somehow managed to get an login error after klicking the googgle sign in button again after I already was logged in with google.

At least we should extend this error page for our base page

image


Not sure if a problem on the autodeploy server, but after successfully login in with OAuth i was always shown this error page (/<flavor>/ E.g /app/ was missing in route) I think on the share there should be a redirect for this anyway, but still

image

I send the buttons to the UX team for them to review.

Oauth feature is working - Nice! - only some finetuning left :)

translations/catroweb.en.yml
dmetzner: the oauthInfo.. where is this information shown? I have not seen it once while testing

leno12: Maybe soemthing like You used an external acccount to log in or something like that? 
To see info you need to clear the cache in the browser. And then you can see it but just the first time when you use some of the oauth services.

src/Catrobat/Security/FOSUBUserProvider.php
leno12: well I will not be looking at this like an ID. Its different. But that it can be very slow thats where I agree. So we need to change this. We can maybe use crc32 for this hashing so then this id would not be so long maybe 8 characters

dmetzner: UX is asking:

can we use the fb/google/apple nickname somehow?
maybe just add (number in case of the same nickname)

leno12: We can use first name and last name. Since in apple implemenation there is no nickname. But nickname is basically users full name so we can use that and then check if somebody already exists with that name and append number in that case

translations/catroweb.en.yml
dmetzner: Yep, external should be fine I guess

dmetzner (approved):

PR #622, opened 2020-06-07
SHARE-314 Extend media lib API

amelak9:
added methods and tests for GET /api/media/files and GET /api/media/file/{id}

### DIFF
--- a/src/Api/MediaLibraryApi.php
+++ b/src/Api/MediaLibraryApi.php
@@ -160,7 +160,25 @@ class MediaLibraryApi implements MediaLibraryApiInterface
    */
   public function mediaFileIdGet(int $id, &$responseCode, array &$responseHeaders)
   {
-    $responseCode = Response::HTTP_I_AM_A_TEAPOT;
+    $json_response_array = [];
+    $media_package_file = $this->entity_manager->getRepository(MediaPackageFile::class)
+      ->findOneBy(['id' => $id])
+        ;
+
+    if (null !== $media_package_file)
+    {
+      $responseCode = Response::HTTP_OK; // 200 => OK
+      $json_response_array[] = new MediaFile($this->getMediaFileResponseData($media_package_file));
+    }
+    else
+    {
+      $responseCode = Response::HTTP_NOT_FOUND; // => Not found
+
+      return null;
+    }
+    $responseData = new MediaFiles(['media_files' => $json_response_array]);
+
+    return $responseData;
   }
 
   /**
@@ -168,6 +186,72 @@ class MediaLibraryApi implements MediaLibraryApiInterface
    */
   public function mediaFilesGet(?int $limit = 20, ?int $offset = 0, ?string $flavor = null, &$responseCode, array &$responseHeaders)
   {
-    $responseCode = Response::HTTP_I_AM_A_TEAPOT;
+    if (null === $limit)
+    {
+      $limit = 20;
+    }
+    if (null === $offset)
+    {
+      $offset = 0;
+    }
+
+    $total_results = 0;
+
+    $json_response_array = [];
+    $media_package_files = null;
+    if (null == $flavor)
+    {
+      $query = $this->entity_manager->createQuery('SELECT u FROM App\Entity\MediaPackageFile u')
+        ->setFirstResult($offset)
+        ->setMaxResults($limit)
+      ;
+      $media_package_files = $query->getResult();
+
+      $num_query = $this->entity_manager->createQuery('SELECT COUNT(u.id) FROM App\Entity\MediaPackageFile u');
+      $total_results = $num_query->getSingleScalarResult();
+    }
+    else
+    {
+      $query = $this->entity_manager->createQuery('SELECT u FROM App\Entity\MediaPackageFile u WHERE u.flavor = :flavor')
+        ->setParameter('flavor', $flavor)
+        ->setFirstResult($offset)
+        ->setMaxResults($limit)
+      ;
+
+      $media_package_files = $query->getResult();
+
+      $num_query = $this->entity_manager->createQuery('SELECT COUNT(u.id) FROM App\Entity\MediaPackageFile u WHERE u.flavor = :flavor')
+        ->setParameter('flavor', $flavor)
+        ;
+      $total_results = $num_query->getSingleScalarResult();
+    }
+    if (null !== $media_package_files)
+    {
+      /** @var MediaPackageFile $media_package_file */
+      foreach ($media_package_files as $media_package_file)
+      {
+        $json_response_array[] = new MediaFile($this->getMediaFileResponseData($media_package_file));
+      }
+    }
+    $responseData = new MediaFiles(['media_files' => $json_response_array, 'total_results' => $total_results]);
+
+    return $responseData;
+  }
+
+  public function getMediaFileResponseData(MediaPackageFile $media_package_file)
+  {
+    return $mediaFile = [
+      'id' => $media_package_file->getId(),
+      'name' => $media_package_file->getName(),
+      'flavor' => $media_package_file->getFlavor(),
+      'package' => $media_package_file->getCategory()->getPackage()->first()->getName(),
+      'category' => $media_package_file->getCategory()->getName(),
+      'author' => $media_package_file->getAuthor(),
+      'extension' => $media_package_file->getExtension(),
+      'download_url' => $this->url_generator->generate(
+          'download_media',
+          ['id' => $media_package_file->getId()],
+          UrlGenerator::ABSOLUTE_URL),
+    ];
   }
 }
--- a/tests/behat/features/api/mediapackage/mediafile_id_get.feature
+++ b/tests/behat/features/api/mediapackage/mediafile_id_get.feature
@@ -1,0 +1,84 @@
+@api @media-package
+Feature: Get data from the media library in json format
+
+  Background:
+    Given there are media packages:
+      | id | name   | name_url |
+      | 1  | Looks  | looks    |
+      | 2  | Sounds | sounds   |
+    And there are media package categories:
+      | id | name    | package |
+      | 1  | Animals | Looks   |
+      | 2  | Fantasy | Sounds  |
+      | 3  | Space   | Looks   |
+
+    And there are media package files:
+      | id | name      | category | extension | active | file   | flavor     | author         |
+      | 1  | Dog 1     | Animals  | png       | 1      | 1.png  | pocketcode | Bob Schmidt    |
+      | 2  | Dog 2     | Fantasy  | mpga      | 1      | 2.mpga | pocketcode |                |
+      | 3  | Spaceship | Space    | png       | 0      | 3.png  |            | Micheal John   |
+      | 4  | Cat       | Animals  | png       | 1      | 4.png  | luna       |                |
+      | 5  | Ape       | Animals  | png       | 1      | 5.png  |            |                |
+      | 6  | Metroid   | Space    | png       | 1      | 6.png  | pocketcode | Jennifer Shawn |
+
+
+  Scenario: Requests with wrong HTTP_ACCEPT value should result in an error
+    Given I have a request header "HTTP_ACCEPT" with value "text/html"
+    And I request "GET" "/api/media/file/5"
+    Then the response status code should be "406"
+
+  Scenario: Requests without specified id should result in an error
+    Given I have a request header "HTTP_ACCEPT" with value "application/json"
+    And I request "GET" "/api/media/file/a"
+    Then the response status code should be "404"
+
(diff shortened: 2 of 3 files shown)


tests/behat/features/api/mediapackage/mediafile_id_get.feature
```diff
@@ -0,0 +1,82 @@
...
+        "author": "",
+        "extension": "png",
+        "download_url": "http:\/\/localhost\/app\/download-media\/5"
+      }
+    ],
+      "total_results": 1
+     }
+    """
+
+  Scenario: Getting file with id should return one media file with that exact id
```
dmetzner, line 63: Why would we need total results here? As far as I understand it, this will always only return 1 result?

tests/behat/features/api/mediapackage/mediafile_id_get.feature
```diff
@@ -0,0 +1,82 @@
...
+        "id": 5,
+        "name": "Ape",
+        "flavor": "pocketcode",
+        "package": "Looks",
+        "category": "Animals",
+        "author": "",
+        "extension": "png",
+        "download_url": "http:\/\/localhost\/app\/download-media\/5"
+      }
+    ],
```
dmetzner: looking at the specification the response should look like this:

image

tests/behat/features/api/mediapackage/mediafile_id_get.feature
```diff
@@ -0,0 +1,82 @@
...
+
+
+  Scenario: Requests with wrong HTTP_ACCEPT value should result in an error
+    Given I have a request header "HTTP_ACCEPT" with value "text/html"
+    And I request "GET" "/api/media/files"
+    Then the response status code should be "406"
+
+  Scenario: Getting file with id should return one media file with that id
+    Given I have a request header "HTTP_ACCEPT" with value "application/json"
+    And I request "GET" "/api/media/file/5"
```
dmetzner, line 42: media/file/id

tests for 400 / 406 are missing

404 is missing in the CAPI specification.
@HCrane you might wanna add this.

tests/behat/features/api/mediapackage/mediafile_id_get.feature
```diff
@@ -0,0 +1,82 @@
...
+      | 2  | Dog 2     | Fantasy  | mpga      | 1      | 2.mpga | pocketcode |                |
+      | 3  | Spaceship | Space    | png       | 0      | 3.png  |            | Micheal John   |
+      | 4  | Cat       | Animals  | png       | 1      | 4.png  | luna       |                |
+      | 5  | Ape       | Animals  | png       | 1      | 5.png  |            |                |
+      | 6  | Metroid   | Space    | png       | 1      | 6.png  | pocketcode | Jennifer Shawn |
+
+
+  Scenario: Requests with wrong HTTP_ACCEPT value should result in an error
+    Given I have a request header "HTTP_ACCEPT" with value "text/html"
+    And I request "GET" "/api/media/files"
```
dmetzner: successful tests are missing (different limit, offset,  flavor values and their defaults)
400 error test is missing

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,96 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+    $media_package_files = $this->entity_manager->getRepository(MediaPackageFile::class)
+      ->findAll()
+        ;
+
+    if (null !== $media_package_files)
+    {
+      /** @var MediaPackageFile $media_package_file */
+      foreach ($media_package_files as $media_package_file)
+      {
+        if ((0 != $offset && count($media_package_files) <= $offset) || count($json_response_array) === $limit)
```
dmetzner: might reduce the compelxity/nesting here + any reason why no strict comparison was applied between 0 != offset?

```
if ((0 !== $offset && count($media_package_files) <= $offset))
{
  $offset -= count($media_package_files);
  continue;
}

if (count($json_response_array) === $limit)
{
  continue;
}
```

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,96 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+    {
+      $responseCode = Response::HTTP_NOT_FOUND; // => Not found
+
+      return null;
+    }
+
+    $responseCode = Response::HTTP_OK; // 200 => OK
+
+    /** @var MediaPackageFile $media_package_file */
+    $mediaFile = [
```
dmetzner: only return the mediafile

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,96 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+      'author' => $media_package_file->getAuthor(),
+      'extension' => $media_package_file->getExtension(),
+      'download_url' => $this->url_generator->generate(
+          'download_media',
+          ['id' => $media_package_file->getId()],
+          UrlGenerator::ABSOLUTE_URL),
+    ];
+    $json_response_array[] = new MediaFile($mediaFile);
+
+    $responseData = new MediaFiles(['media_files' => $json_response_array, 'total_results' => 1]);
```
dmetzner: only return the mediafile and remove the rest

dmetzner (changes requested):

src/Api/MediaLibraryApi.php
```diff
@@ -233,25 +219,29 @@ public function mediaFilesGet(?int $limit = 20, ?int $offset = 0, ?string $flavo
...
-        $json_response_array[] = new MediaFile($mediaFile);
+        $json_response_array[] = new MediaFile($this->addMediaFile($media_package_file));
       }
     }
     $responseData = new MediaFiles(['media_files' => $json_response_array, 'total_results' => $total_results]);
 
     return $responseData;
   }
+
+  public function addMediaFile(MediaPackageFile $media_package_file)
```
dmetzner: Since this method is not adding a new MediaFile, it would be better to rename it so something like `getMediaFileResponseData`

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,98 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+    }
+    else
+    {
+      $media_package_files = $this->entity_manager->getRepository(MediaPackageFile::class)
+        ->findBy(['flavor' => $flavor])
+        ;
+    }
+    if (null !== $media_package_files)
+    {
+      $total_results = count($media_package_files);
```
dmetzner: this in combination with all data (1) is very very slow on a large dataset.
Use 2 different queries.

- One to get the data (the query supports and can already use limit + offset ~ no need to reimplement the limit offset logic)
- The second to get the number of total results. Important the query builder should not load all the results only the number (Something like getScalarResults)

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,98 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+    }
+
+    $total_results = 0;
+
+    $json_response_array = [];
+    $media_package_files = null;
+    if (null == $flavor)
+    {
+      $media_package_files = $this->entity_manager->getRepository(MediaPackageFile::class)
+        ->findAll()
```
dmetzner: (1)

dmetzner (changes requested):
- First of all, please **squash your changes**. The commit history looks kinda broken. I'm a co-author in your git history!?
`
git reset --soft HEAD~X       
`                       
`X` is the number of commits to remove (changes will not be deleted due to the "soft" option)
`
git commit -a                                              
`
Enter Your commit message and force push

- Remove the package-lock changes. 

- total results have a major performance problem in the current implementation. Please use queries with limit and offset directly and create a second query for the number of total results. You might want to take a look at the projects Api.

- The tests look fine to me. But if you take a look at e.g. the project api tests. @JackShahhoud refactored them in a way to get rid of the long json strings in the tests. Something you could reconsider for the MediaLib too.  (E.g. `tests/behat/features/api/projects/get_random_projects.feature`)

dmetzner:
This might helps
`public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $offset = null);`

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,102 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+    }
+
+    $total_results = 0;
+
+    $json_response_array = [];
+    $media_package_files = null;
+    if (null == $flavor)
+    {
+      $query = $this->entity_manager->createQuery('SELECT u FROM App\Entity\MediaPackageFile u');
+      $media_package_files = $query->getResult();
```
dmetzner, line 208: You are still loading **ALL** results here. On share, this can be multiple thousands of assets.

**Option 1: DQL** 

Why? Easy to read if we know the whole query. And that's the case. There will never be a query without limit/offset here.

```
$query = $this->entity_manager->createQuery('SELECT u FROM App\Entity\MediaPackageFile u')
->setFirstResult($offset)
->setMaxResults($limit);
```

dmetzner: **Option 2: QueryBuilder**

Why?: Easier to build complex queries with conditions and loops if we do not know the whole query.

Take a look for example at the projects API.
Making use of the methods defined in the ProgramRepository.

```
public function getMostViewedPrograms(bool $debug_build, string $flavor = null, int $limit = 20, int $offset = 0,
                                        string $max_version = '0'): array
  {
    $query_builder = $this->createQueryBuilder('e');

    $query_builder
      ->select('e')
      ->where($query_builder->expr()->eq('e.visible', $query_builder->expr()->literal(true)))
      ->orderBy('e.views', 'DESC')
      ->setFirstResult($offset)
      ->setMaxResults($limit)
    ;

    $query_builder = $this->addPrivacyCheckCondition($query_builder);
    $query_builder = $this->addDebugBuildCondition($query_builder, $debug_build);
    $query_builder = $this->addFlavorCondition($query_builder, $flavor);
    $query_builder = $this->addMaxVersionCondition($query_builder, $max_version);

    return $query_builder->getQuery()->getResult();
  }
```

dmetzner: You could extend the Media Repositories to your need.
- `MediaPackageFileRepository.php`
- `MediaPackageRepository.php`
- `MediaPackageCategoryRepository.php`

A small sidenote: 
For some unknown reason the file is in
`src/Catrobat/Services/MediaPackageFileRepository.php`

Feel free to move it to `src\Repository` where it belongs.

dmetzner: **Option 3: For the sake of completeness**

You also directly append `LIMIT` and `OFFSET` to the query creation. But ensure you bind the values correctly in order to prevent sql injections.

`setFirstResult, setMaxResults, setParameter `do that all ready for you. Might wanna check out https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/security.html#user-input-and-doctrine-orm

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,102 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+        ->setParameter('flavor', $flavor)
+        ;
+      $total_results = $num_query->getSingleScalarResult();
+    }
+    if (null !== $media_package_files)
+    {
+      /** @var MediaPackageFile $media_package_file */
+      foreach ($media_package_files as $media_package_file)
+      {
+        if (0 != $offset)
```
dmetzner: we can get rid of both ifs here then

dmetzner (changes requested):

src/Api/MediaLibraryApi.php
```diff
@@ -160,14 +160,102 @@ public function mediaPackagePackageNameGet(string $package_name, ?int $limit = 2
...
+    {
+      $query = $this->entity_manager->createQuery('SELECT u FROM App\Entity\MediaPackageFile u');
+      $media_package_files = $query->getResult();
+
+      $num_query = $this->entity_manager->createQuery('SELECT COUNT(u.id) FROM App\Entity\MediaPackageFile u');
+      $total_results = $num_query->getSingleScalarResult();
+    }
+    else
+    {
+      $query = $this->entity_manager->createQuery('SELECT u FROM App\Entity\MediaPackageFile u WHERE u.flavor = :flavor')
```
dmetzner, line 215: If you go if Option 2:

This flavor condition could be dynamically added to the query only if needed, heavily reducing the duplicated code

dmetzner (approved):
For future contributions try to keep your commit message up-to-date and meaningful.

```
Ticket number + title

what changed (and if necessary why)
```

PR #662, opened 2020-06-17
SHARE-154 Detail page for Scratch projects

11possim:
Added the corresponding scratch IDs to program and user entities.
Added /scratch_project/{id} endpoint, it creates a catrobat project (if there isn't one already) and redirects to that project.
A user is also added (if it doesn't already exist). The usernames for Scratch users are: 'Scratch:<Name>'.
Added a 'Profile on Scratch' Button in profile view of scratch users.
Didn't change program frontend b/c currently in redesign (Share-267).
Remix relations aren't added at the moment.
Added the /scratch_project/{id} endpoint to robots.txt to prevent the creation of new projects by crawler.

---

### Additional Description


dmetzner:
It works like a charm! ~Even the download, which was not required by the ticket, works.~

- ~**missing indication** that it's a scratch project other than the username. 
` The detail page should include a static, translated text right after the description, that mentions that this is a Scratch project, and includes the link to it.` I guess also because of  (Share-267)?
At least we should **create a new ticket** for it. Else we will forget about it.~

- The **Scratch User profile button** ~looks a bit weird. I'll contact UX and ask them before I say something wrong:~ Please add some spacing in between for the mobile view. The button will find it's final position in the profile redesign. (UX updates the mockups)

image

## Some thoughts for discussion:
@11possim @bernadettespieler @HCrane 

- `/scratch_project` route - should we maybe change it to `/scratch/project` to be more compatible with possible future additions? like `/scratch/user` to import whole users from scratch 

- the logic behind `/scratch_project` must be added to the CatrowebAPI. **create a new ticket?**

- ~should we prevent username creation with a colon sign? So only this autogenerated usernames can have the `Scratch:` prefix. Else I could create an account 'Scratch:ABC" and "steal" ABC identity/projects.~ **Update Capi Documentation**

- `From the ticket description: "The links to the detail pages of Scratch projects on our server should NOT be shown in any section of the main page, i.e. not in the “Newest projects” section etc."`
   Why would we want this? After the conversion, the project is basically is a Catrobat project. If it is loved by many people or a cool game, why shouldn't we recommend it to others? Plus isn't the scratch category on the landing page explicitly for such projects?

- ~What happens if the scratch user updates his project on scratch? Can we somehow check out the new updated version of the project~ The updated question - when and how can this conversion be done? Users will never visit the  `/scratch_project` URL. How can they convert projects?  (happens this only during the search?")

- ~should we also "steal" the likes/reactions, views, downloads, comments from scratch projects, and add them to our catrobat version project?~

src/Api/UserApi.php
```diff
@@ -124,6 +124,10 @@ public function validate(Register $register): ValidationSchema
     {
       $validation_schema->setUsername($this->translator->trans('api.registerUser.usernameAlreadyInUse', [], 'catroweb'));
     }
+    elseif (0 === strncmp($register->getUsername(), User::$SCRATCH_PREFIX, strlen(User::$SCRATCH_PREFIX)))
```
dmetzner: AH nice, you already thought about this! :)
So, we just need to update the CAPI documentation

src/Entity/ProgramManager.php
```diff
@@ -307,6 +300,80 @@ public function addProgram(AddProgramRequest $request): ?Program
...
+    }
+    else
+    {
+      if ($program->getLastModifiedAt() === $modified_time)
+      {
+        return $program;
+      }
+      $program->incrementVersion();
+    }
+    $program->setCatrobatVersion(1);
```
dmetzner, line 325: I'm not sure about this version number. Hardcoding it will result in many problems. Catblocks + IOS team depend on this number.

src/Entity/ProgramManager.php
```diff
@@ -307,6 +300,80 @@ public function addProgram(AddProgramRequest $request): ?Program
...
+    $program->setVisible(true);
+    $program->setApproved(false);
+    $program->setCatrobatVersion(1);
```
dmetzner: + why set it twice ;)

src/Catrobat/Services/OAuthService.php
```diff
@@ -343,7 +343,7 @@ public function loginWithTokenAndRedirectAction(Request $request): JsonResponse
     return JsonResponse::create($retArray);
   }
 
-  public function randomPassword(): string
+  public static function randomPassword(): string
```
dmetzner: we should move this to some utils

tests/behat/features/web/project-details-3/project_scratch.feature
```diff
@@ -0,0 +1,32 @@
...
+Feature: As a visitor I want to see scratch projects
+
+  Scenario: Visit a Scratch program
+    Given I request "GET" "/app/scratch_project/401650420"
+    Then I should be redirected to a catrobat program
+    And there should be "1" programs in the database
+    And there should be "1" users in the database
+
+  Scenario: Visiting a not existing Scratch program
+    #Given I request "GET" "/app/scratch_project/1"
```
dmetzner: unused code (comment) which can be removed

dmetzner (changes requested):
You also have changed almost all shell scripts. Any reason for this "0 lines, but bytes have changed" changes?

11possim:
> Even the download, which was not required by the ticket, works.

Just checked, that's a bug. [(SHARE-332)](https://jira.catrob.at/browse/SHARE-332)

> should we also "steal" the likes/reactions, views, downloads, comments from scratch projects, and add them to our catrobat version 

I only found a deprecated endpoint for the comments (e.g.: https://scratch.mit.edu/site-api/comments/project/142/). If we want to add the comments we would have to check the request their client makes.
Scratch allows to comment on other comments, share doesn't.
Some scratch projects have a lot of views & downloads, those sections on the homepage would then only show scratch projects, and it would also possibly skew with the recommendation algorithm.
I don't think you can see by whom something is liked in scratch (same for favourites).

> What happens if the scratch user updates his project on scratch? Can we somehow check out the new updated version of the project

It would be possible to check if a project is a scratch project when it gets opened with the /project/ url and then check for updates.
Another possibility would be to add a command that checks all scratch projects for updates. (A problem there would be the rate limit of the Scratch API)

> You also have changed almost all shell scripts. Any reason for this "0 lines, but bytes have changed" changes?

The import-container-libs script had CRLF line separators, so it didn't work with linux hosts. Some other scripts didn't have the executable flag.

PS: at least in the old version of the Scratch API 10 request per second were allowed to be made and returned at max 40 results(projects, comments, ...) at once.

src/Entity/ProgramManager.php
11possim: All usages of setCatrobatVersion have 1 as parameter. Maybe we should delete that column?
The info about the catrobat version seems to be in CatrobatVersionName.
Should that one be set even if no project file currently exists?

dmetzner: Yeah, for sure, let's delete it. But it can be done on a different ticket. It would be necessary to rework this version feature anyway. Instead of the string should use the typical major.minor.patch format. The sql string comparison right now is a little bit hacky - requiring exactly the correct format (same string length) to work. **Ticket needed**

dmetzner: No if there is no project it should not be set.

dmetzner:
> It would be possible to check if a project is a scratch project when it gets opened with the /project/ url and then check for updates.

I would prefer this solution. The project can be updated in the background if necessary + no manual commands or cron jobs are needed.

src/Entity/ProgramManager.php
11possim: Added [SHARE-333](https://jira.catrob.at/browse/SHARE-333)

11possim:
Added ScratchController & ScratchManager, moved /scratch_project/{id} to /scratch/project/{id}, allow POST Requests
Added async updating of project when a Scratch project is visited
Implemented requested changes

dmetzner (approved):
LGTM

PR #722, opened 2020-07-10
SHARE-346 API Search users

JackShahhoud:
-Implement Search users
-Write tests

---

- [x] Include the name and id of the Jira ticket in the PR’s title

### DIFF
--- a/tests/behat/features/api/user/search_users.feature
+++ b/tests/behat/features/api/user/search_users.feature
@@ -1,0 +1,81 @@
+@api @user
+Feature: Search users
+
+  To find users, users should be able to search all available users for specific words
+
+  Background:
+    Given there are users:
+      | name         | password | token      | id |
+      | Catrobat     | 12345    | cccccccccc | 1  |
+      | User1        | vwxyz    | aaaaaaaaaa | 2  |
+      | NewUser      | 54321    | bbbbbbbbbb | 3  |
+      | Catroweb     | 54321    | bbbbbbbbbb | 4  |
+      | пользователь | 54321    | bbbbbbbbbb | 5  |
+    And there are followers:
+      | name     | following      |
+      | Catrobat | User1, Catroweb|
+      | NewUser  | Catrobat       |
+      | Catroweb | User1, NewUser |
+    And there are programs:
+      | id        | name            | owned by | version | private | visible |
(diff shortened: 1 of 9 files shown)


tests/behat/features/api/user/search_users.feature
```diff
@@ -0,0 +1,68 @@
...
+
+    Given I have a parameter "query" with value "Catroweb"
+    And I have a request header "HTTP_ACCEPT" with value "application/json"
+    And I request "GET" "/api/users/search"
+    Then the response status code should be "200"
+    Then the response should have the users model structure
+    Then the response should contain users in the following order:
+      | Name      |
+      | Catroweb  |
+
```
dmetzner, line 80: we have many russian users, a test with russian chars would be nice :)

dmetzner (changes requested):
-  maybe add the russian character testcase.

- Concerning the image... we should discuss this with the teams. But definetly we should only use one variant. Url or base64 image - but not both. @HCrane 

 If we use the url variant -> clear/small api response, but second call needed to get the image + a rework of the profile image system is necessary or at least an additional api. (Similar to github downlaod artifacts)

if we use base 64 variant -> not much to do, just instead of the default image return 
a) the default image in base 64 format
b) return an empty string
+ we slightly need to adapt the description on Catroweb-API repo

- the search itself is interesting - maybe you can explain it:
```

query = 'a' => returns [ Angela ]

query = 'catro' =>return [ catroweb ]

query = 'c' => return [ all users .. i guess because of the localhost email?]   

query = 'o' => returns nothing... so maybe not the localhost mail address...

query = 'atroweb' returns nothing .. why is the 'c' so important here?

query = 'ca' returns [ catroweb ]
```

is it possible to reduce the importance of the 1st char?

rest looks good to me

tests/behat/features/api/user/search_users.feature
JackShahhoud: Good Idea, although i don't speak Russian, i will try to write tests using Russian alphabet. @dmetzner

HCrane: Google Translate is our friend! Dosen't have to be meaningfull names just russian chars!

dmetzner (approved):

PR #741, opened 2020-07-20
SHARE-353 featured projects bug

dmetzner:
- a project must only be either approved or featured to be still visibile
  - count must be a int

#### Additional information

Informed other teams/admins about how they must use the feature.


HCrane (approved):
LGTM

PR #779, opened 2020-07-28
SHARE-255/Refactoring - Adminarea - Tools/Maintain

amelak9:


tests/behat/features/web/admin/server_maintenance.feature
```diff
@@ -14,6 +14,9 @@ Feature: Admin Server Maintenance
     When I GET "/admin/maintain/list"
     Then the client response should contain "Delete APKs"
     And the client response should contain "Delete extracted files"
+    And the client response should contain "Delete log files"
+    And the client response should contain "Archive logs files"
+
 
   Scenario: As a valid admin I want to be able to remove the APKs through the backend
```
dmetzner, line 21: Would be nice to have tests for this section
image

dmetzner (changes requested):
More tests please - unit would be better but integration tests are also okay.
Try to get a higher coverage in those files:
image


and if possible also for the admin maintain controller

dmetzner (approved):

PR #785, opened 2020-07-30
SHARE-365 Extensions as flavor

dmetzner:
Use extension tags to find projects similar to flavor

### DIFF
--- a/src/Catrobat/Services/TestEnv/DataFixtures/ProjectDataFixtures.php
+++ b/src/Catrobat/Services/TestEnv/DataFixtures/ProjectDataFixtures.php
@@ -101,8 +101,8 @@ class ProjectDataFixtures
     );
     $project->setRemixMigratedAt(null);
     $project->setCatrobatVersion(1);
-    $project->setCatrobatVersionName($config['version'] ?? '');
-    $project->setLanguageVersion($config['language version'] ?? '0.92');
+    $project->setCatrobatVersionName($config['version'] ?? '0.8.5');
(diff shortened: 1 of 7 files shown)


leno12 (changes requested):
PR looks nice to me. Just if you can fix the failing test cases it would be great.

src/Catrobat/Services/TestEnv/DataFixtures/ProjectDataFixtures.php
```diff
@@ -101,8 +101,8 @@ public function insertProject(array $config, bool $andFlush = true): Program
     );
     $project->setRemixMigratedAt(null);
     $project->setCatrobatVersion(1);
-    $project->setCatrobatVersionName($config['version'] ?? '');
-    $project->setLanguageVersion($config['language version'] ?? '0.92');
+    $project->setCatrobatVersionName($config['version'] ?? '0.8.5');
+    $project->setLanguageVersion($config['language version'] ?? '0.925');
```
HCrane, line 105: This needs further investigation!

HCrane (approved):
Primarily LGTM but this Version thing needs further ivestigation...@dmetzner would you be so kind to create a research ticket?

src/Catrobat/Services/TestEnv/DataFixtures/ProjectDataFixtures.php
dmetzner: Done

PR #793, opened 2020-08-01
SHARE-276 Redesign project list (grid, pagination ...) Part 2: add\fix tests

dmetzner:
Extend https://github.com/Catrobat/Catroweb/pull/661 by @mfuerni 

 - PocketCode flavor uses projects from all flavors
 - fix routing of project list (test-env)
 - new project_list tests
 - adapted old tests
 - disabled some tests and created tickets to rework those features (recommendations & click stats)

There are chrome exceptions in behat and mink if too many features are excecuted in on suite. The cache must be cleared manually before the failing test can be executed again in order to achieve a success. Without clearing the cache the test will always result in a chrome exception. Nevertheless without running the other tests before the one with the chrome exception is no problem. So the problem is hard to identify... Simple solution for now ... more suites


HCrane (approved):
👍

PR #1029, opened 2020-10-29
HOTFIX Regex

HCrane:
- iOS Projects not properly stored --> fixed by updating the Regex because different XML versions


dmetzner (approved):
Can't say anything about all possible Schemas with a different fileName version that could be possible out there in the wild.

However, it works for the (new one?)

```
...
<look fileName="261575712134352_imageName.png" name="sky"/>
...
```

and also for the(old?)  iOS one:


```
...
<fileName>1f363a1435a9497852285dbfa82b74e4_imageName.png</fileName>
...
```

PR #1785, opened 2021-07-01
Translate Program Section

JoshuaSchaefer:
- Add API call when translate button is selected in program section
- Replace title, description, and credits if available with translated text
- Display translated by line with translation provider, source language and target language
- If API fails, the user will be redirected to google translate

---

- [ ] Include the name and id of the Jira ticket in the PR’s title eg.: `SHARE-666 The devils ticket`
- [x] Verify that all tests are passing (CI), if not please state the test cases in the [section](#Tests) below
- [ ] Put your ticket into the `Code Review` section in [Jira](https://jira.catrob.at/)
- [x] Check that your pull request has been successfully deployed to https://web-test-1.catrob.at/

### Additional Description


assets/js/custom/TranslateComments.js
```diff
@@ -0,0 +1,79 @@
+/* eslint-env jquery */
+/* global Translation */
+/* global ByLineElementContainer */
+
+// eslint-disable-next-line no-unused-vars
+class TranslateComments extends Translation {
```
JoshuaSchaefer, line 6: The TranslateComments class handles the translation function specific to the comment section. It contains listeners for each of the translate and remove translation buttons for each comment. Then, it also contains methods that set, open, and close the translated content within each comment.

assets/js/custom/TranslateProgram.js
```diff
@@ -0,0 +1,135 @@
+/* eslint-env jquery */
+/* global Translation */
+/* global ByLineElementContainer */
+
+// eslint-disable-next-line no-unused-vars
+class TranslateProgram extends Translation {
```
JoshuaSchaefer, line 6: The TranslateProgram class is similar to the TranslateComments class, but handles the translation function specific to the program section. It contains listeners for the translate and remove translation button for the program. Then it contains methods that set, open, and close the translations for the title, description, and credits of the project.

assets/js/custom/Translation.js
```diff
@@ -1,42 +1,22 @@
...
-    const commentId = $(this).attr('id').substring('comment-translation-button-'.length)
-
-    $(this).hide()
-
-    if (isTranslationNotAvailable(commentId)) {
-      $('#comment-translation-loading-spinner-' + commentId).show()
-      translateComment(commentId)
-    } else {
-      openTranslatedComment(commentId)
+class Translation {
```
JoshuaSchaefer, line 4: To help split up the program translation and comment translation functionality, the Translate function has now been replaced with a Translation base class. The new Translation class contains the same functionality as the previous Translation function by getting and splitting up the translatedByLine, then also retrieving all of the display languages to show in the translatedByLine. It also contains a new method to set the translatedByLine for both the program and comment section. However, the functions that set, opened, and removed the translations from the comment section have now been moved to the TranlsateComment class which extends this new Translation class.

templates/Program/program.html.twig
```diff
@@ -27,12 +27,27 @@
     </div>
 
     <div class="col-7 col-sm-8 col-md-8 col-lg-5">
-      <div style="display: flex;">
-        <span id="name" class="h2" style="min-width: 30px; white-space: nowrap; text-overflow: ellipsis; overflow: hidden;">{{ program.name }}</span>
-        <span style="flex: 1; white-space: nowrap;">
-          <a class="icon-text align-bottom" id="translate-program">
-            <i class="ml-2 material-icons">translate</i>
-          </a>
+      <div style="display: inline-flex;">
```
dmetzner: max-width: 100% is missing.

Without max-width set to 100% the whole page is a side scroller on large names
image


With max-width fixed to 100%
image


In Addition, move those styles to a class :)

dmetzner (changes requested):
Something I noticed. Most text phrases are very small and iTranslate has a very hard job distinguishing the correct language (E.g. languages such as russian vs serbian). However, that is nothing we could change easily. Maybe we could store some metadata in the project that allows users that are uploading a project to set their language. But only in the future, if necessary and required by the feedback of our users. All in all, I love this feature - just a minor css issue that should be fixed! :) :+1:

templates/Program/program.html.twig
JoshuaSchaefer: I created a new name-wrapper class with max-width set to 100% and the display style set to inline-flex. I think in the next few weeks we will be working on adding an optional manual translation where program owners can define their own translations. I will try to add a place that users can set the language of their original content to improve the performance of the API within those changes for the manual translation. That's a good idea and a good catch on the styling 😀

dmetzner (approved):

danielzhang130:
> Something I noticed. Most text phrases are very small and iTranslate has a very hard job distinguishing the correct language (E.g. languages such as russian vs serbian). However, that is nothing we could change easily. Maybe we could store some metadata in the project that allows users that are uploading a project to set their language. But only in the future, if necessary and required by the feedback of our users. All in all, I love this feature - just a minor css issue that should be fixed! :) 👍

That's an issue we need to think about in the future.

Couple of ideas:
- If we store the translation results, including the source language detected by the API, we might be able to infer statistically the most likely correct language.
- We might be able to store the display language for the user in their profile and use that as the source language for their uploaded programs.
- how different are Serbian and Russian? "Si" often has the same meaning in French and Spanish, so it wouldn't affect too much the translation results.
- (save the best for last) google translate might be better that this? Also Google's quota is character quota, and itranslate is API count, so it make sense to use Google for shorter request and itranslate for long text.

dmetzner:
> how different are Serbian and Russian? 

Not my area. So I better don't say anything to this. It would only be a random guess :)


I like the ideas, however, as I have already mentioned I guess it is not that important to implement such minor changes/improvements immediately. Let's wait for the feedback of your users. :)

PR #3929, opened 2023-04-17
SHARE-535 Feature flag management for new features

EifonUser:
---

- [ ] Include the name and id of the Jira ticket in the PR’s title eg.: `SHARE-666 The devils ticket`
- [ ] Choose the proper base branch (*develop*)
- [ ] Confirm that the changes follow the project’s coding guidelines
- [ ] Verify that the changes generate no warnings and errors 
- [ ] Verify to commit no other files than the intentionally changed ones
- [ ] Include reasonable and readable tests verifying the added or changed behavior
- [ ] Verify that all tests are passing (CI), if not please state the test cases in the [section](#Tests) below
- [ ] Perform a self-review of the changes
- [ ] Stick to the project’s git workflow (rebase and squash your commits)
- [ ] Verify that your changes do not have any conflicts with the base branch
- [ ] Put your ticket into the `Code Review` section in [Jira](https://jira.catrob.at/)
- [ ] Check that your pull request has been successfully deployed to https://web-test-1.catrob.at/

### Additional Description


migrations/2023/Version20230417220111.php
```diff
@@ -0,0 +1,41 @@
...
+  public function getDescription(): string
+  {
+    return '';
+  }
+
+  public function up(Schema $schema): void
+  {
+    // this up() migration is auto-generated, please modify it to your needs
+    $this->addSql('CREATE TABLE FeatureFlag (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(255) NOT NULL, value TINYINT(1) NOT NULL, PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB');
+    $this->addSql('DROP TABLE feature_flag');
```
dmetzner: Create + drop? 
I think you got your final migration mixed up with a previous state of your database

migrations/2023/Version20230417220111.php
```diff
@@ -0,0 +1,41 @@
...
+    $this->addSql('ALTER TABLE studio_activity CHANGE type type ENUM(\'comment\', \'project\', \'user\')');
+    $this->addSql('ALTER TABLE studio_user CHANGE role role ENUM(\'admin\', \'member\'), CHANGE status status ENUM(\'active\', \'banned\', \'pending_request\')');
+    $this->addSql('ALTER TABLE user_like_similarity_relation CHANGE similarity similarity NUMERIC(4, 3) DEFAULT \'0\' NOT NULL');
+    $this->addSql('ALTER TABLE user_remix_similarity_relation CHANGE similarity similarity NUMERIC(4, 3) DEFAULT \'0\' NOT NULL');
```
dmetzner: FYI: those lines are currently generated at all times :/
It's a bug in the migration diff command in combination with our Entites & DB. You can just delete them before committing the migration file. On the other hand this results in no issues... just leaves the file with unnecessary content

src/Admin/Tools/FeatureFlag/FeatureFlagManager.php
```diff
@@ -0,0 +1,77 @@
...
+
+class FeatureFlagManager
+{
+  private array $defaultFlags;
+
+  public function __construct(
+    protected RequestStack $requestStack,
+    protected EntityManagerInterface $entityManager
+  ) {
+    $this->defaultFlags = include __DIR__.'/../../../../config/features.php';
```
dmetzner: symfony has the parameter bag interface, which allows you to get variables such as the root directory; this would make the path more readable than stacking ../

dmetzner (changes requested):

src/Admin/Tools/FeatureFlag/FeatureFlagManager.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    $entityManagerFlags = $this->entityManager->getRepository(FeatureFlag::class)->findAll();
+    foreach ($entityManagerFlags as $flag) {
+      $flagName = $flag->getName();
+      if (!array_key_exists($flagName, $flagMap)) {
+        $this->entityManager->remove($flag);
+      }
+    }
+    $this->entityManager->flush();
```
dmetzner, line 39: Something to consider - however, better done in a new PR

Checking the DB multiple times in the constructor might not be the best idea.
If you have, for example, five feature flags, you get at least six more DB queries every time this manager is used, even if no create/delete queries are needed.

You could reduce the needed amount of queries by first querying for flags in the DB (A), and also by reading all from the configuration (B)

Run a batch delete for all flags found in A, but not in B 
Create each flags in B not found in A

In the best case, you only need 1 query instead of 6. and also optimized the deletion process.

Nonetheless, this "syncing" process should not be done in the constructor. Consider moving it to the Special Updater, which is executed only once at every deployment, leaving us with the following feature flag in the DB

id=auto
name=loadedFromConfig
configValue=loadedFromConfig
value=null/true/false, where null means it was not set in the admin interface, and we just need the config value.

while the config value is updated on every deployment, the value should be kept the same.

This basically means during requests by our users we never have to check a config file (reading the filesystem is slow) and also only need a single query, either we have the value set, then use that value, and if not we use the config value (provided by the DB object) This check can be hidden to the developer by using the manager as a facade. 

So on a request to a feature with a feature flag, we check the header (cheap & fast), and if no value is set in the header, we call our facade (the manager), which costs at max. 1 DB query. To spin this even further, we could cache those results in the Symfony cache for e.g. 1 day to even get rid of this query

EifonUser:
I dont know what happend with the database. Thank you for pointing that out. Also paramterbag is now being used for the path of features.php

schaubes (approved):

maxfazeli-dev (approved):

dmetzner (approved):

PR #5057, opened 2024-07-03
SHARE-595 Introduction Task - Edmond

Edmondnrecaj:
Behat tests are missing, still need to get more familiar with them.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # Use composer image for better caching
-FROM composer:2.7.7 AS composer-build
+FROM composer:2.7.8 AS composer-build
 COPY composer.json composer.lock ./
 RUN composer install --no-scripts
 


templates/Project/project_details.html.twig
```diff
@@ -41,6 +41,29 @@
...
+        {% include 'components/redirect_button.twig' with {
+        'id': 'projectStealButton',
+        'url': path('project_steal', {'id': project.id }),
+        'icon': 'person',
+        'text': "Steal"
+      } %}
+      <aside class="mdc-snackbar">
+        <div class="mdc-snackbar__surface" role="status" aria-relevant="additions">
+          <div class="mdc-snackbar__label" aria-atomic="false">
+            Can't send photo. Retry in 5 seconds.
```
dmetzner: Since you have not used the translation system, just a heads up for the future:

For example `codeview.title` is registered by the developer in the `translations/catroweb.en.yaml` file as 
```
codeview:
    title: Steal
```

All other translation files; E.g `catroweb.de.yml` are translated in its own process (automated by GitHub Actions) using Crowdin -> More details here: https://catrobat.atlassian.net/wiki/spaces/CATWEB/pages/74383377/Crowdin+-+Translation+Management

src/Application/Controller/Project/ProjectController.php
```diff
@@ -212,6 +212,26 @@ public function projectLike(Request $request, string $id): Response
     ]);
   }
 
+  /**
+   * @throws NoResultException
+   */
+  #[Route(path: '/project/steal/{id}', name: 'project_steal', methods: ['GET'])]
```
dmetzner: At Catroweb we try to write our code using the API approach, to decouple the frontend from the backend. We have multiple different frontends (ours on the share, and also the apps)
Usually, your request (steal) would use our API (https://github.com/Catrobat/Catroweb-API // https://developer.catrobat.org/Catroweb-API/) Of course that is an oerhead for the training example.

Instead of using a GET request, consider using POST which is used to send data to a server to create/update a resource. Feel free to take a look at all available methods: https://www.w3schools.com/tags/ref_httpmethods.asp

That said it would be nice, to return a status code rather than a redirect in the response, and handle the response using the frontend logic.

src/Application/Controller/Project/ProjectController.php
```diff
@@ -212,6 +212,26 @@ public function projectLike(Request $request, string $id): Response
...
+  public function projectSteal(Request $request, string $id): Response
+  {
+    $project = $this->project_manager->find($id);
+    $currentuser = $this->getUser();
+    $project->setUser($currentuser);
```
dmetzner: what would happen here if the project does not exist?

dmetzner:
>1) src/Application/Controller/Project/ProjectController.php (single_quote, braces_position, blank_line_before_statement)
Found 1 of 903 files that can be fixed in 12.919 seconds, 20.00 MB memory used
Error: Process completed with exit code 8.

With `php composer.phar fix` those code style issues can be resolved automatically

Issue #6299, opened 2026-03-18
Bug: Follower/Following Logic Swap in API and Trust Score Calculation


abhinavsudhik:
**Describe the Catroweb bug**
The follower/following logic is inverted in two locations in the codebase, which causes the Followers API to return incorrect data and trust scores to be calculated with the wrong follower count.

1. `src/Api/Services/Followers/FollowersApiLoader.php` — `getFollowers()` queries `f.following` (should be `f.followers`) and `getFollowing()` queries `f.followers` (should be `f.following`)
2. `src/Moderation/TrustScoreCalculator.php` (lines 92–100) — `$follower_count` joins on `f.following`, counting who the user follows rather than who follows them

**To Reproduce**
1. Go to any user profile on share.catrob.at with an asymmetric follower/following count
2. Check the follower and following lists
3. Observe that the two lists are swapped — followers shows who the user follows, and vice versa

**Expected behavior**
- The followers list should show users who follow the given user
- The following list should show users the given user follows
- Trust score calculation should use the correct incoming follower count

**Screenshots**
N/A — reproducible via API response or profile page with asymmetric follower/following counts

**Desktop/Smartphone:**
- Device: PC
- OS: N/A (server-side bug)
- Browser: Any
- Version: N/A

**Additional context**
Affected files and lines:
- `src/Api/Services/Followers/FollowersApiLoader.php` lines 23–24, 38–39
- `src/Moderation/TrustScoreCalculator.php` lines 92–100

The fix is to swap the `f.following` / `f.followers` references in these three locations so each query joins and counts the correct side of the relationship. Moderation trust scores may currently be incorrect for users where follower and following counts differ significantly.

V-inaya-K:
Hi! I’d like to work on this issue for GSoC 2026. I’ll start exploring and submit a PR soon 🙂

dmetzner:
After thorough analysis, the current code is **correct** — the follower/following logic is not inverted. Closing as not-a-bug.

### Detailed explanation

The User entity defines a self-referencing ManyToMany:

```php
#[ManyToMany(targetEntity: User::class, mappedBy: 'following')]
protected Collection $followers;   // inverse side

#[ManyToMany(targetEntity: User::class, inversedBy: 'followers')]
protected Collection $following;   // owning side
```

When user A follows user B, `$a->addFollowing($b)` is called (see `FollowersApiProcessor::followUser()`), so `A.following` contains B and `B.followers` contains A.

[...]

PR #6302, opened 2026-03-18
fix follower/following inversion bug #6299

V-inaya-K:
This PR fixes incorrect follower/following logic in the codebase.

Changes:

* Corrected usage of `f.followers` and `f.following` in FollowersApiLoader.php
* Fixed trust score calculation in TrustScoreCalculator.php

Result:

* Followers API now returns correct data
* Trust scores are calculated using the correct follower count

### DIFF
--- a/src/Api/Services/Followers/FollowersApiLoader.php
+++ b/src/Api/Services/Followers/FollowersApiLoader.php
@@ -20,8 +20,8 @@ class FollowersApiLoader extends AbstractApiLoader
    */
   public function getFollowers(User $user): array
   {
-    $followers = $this->queryRelatedUsers($user, 'f.following');
-    $total_following = $this->countRelatedUsers($user, 'f.followers');
+    $followers = $this->queryRelatedUsers($user, 'f.followers');
+    $total_following = $this->countRelatedUsers($user, 'f.following');
 
     return [
       'users' => $followers,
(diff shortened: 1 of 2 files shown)


dmetzner:
This PR should not be merged — the current code is already correct, and this swap would **introduce** the bug it claims to fix.

### Why the current code is correct

The User entity defines:

```php
#[ORM\ManyToMany(targetEntity: User::class, mappedBy: 'following')]
protected Collection $followers;   // inverse side

#[ORM\ManyToMany(targetEntity: User::class, inversedBy: 'followers')]
protected Collection $following;   // owning side
```

When user A follows user B, `$a->addFollowing($b)` is called (see `FollowersApiProcessor::followUser()`). So `A.following` contains B, and `B.followers` contains A.

The query in `queryRelatedUsers` is:

```php
FROM User f JOIN {$joinRelation} u WHERE u.id = :targetUserId
```

- **`JOIN f.following u WHERE u.id = :X`** → finds all `f` who have X in their `following` → `f` follows X → these are **X's followers** ✅
- **`JOIN f.followers u WHERE u.id = :X`** → finds all `f` who have X in their `followers` → X follows `f` → these are **people X follows** ✅

So the current mapping is:

| Method | Join relation | Returns | Correct? |
|---|---|---|---|
| `getFollowers()` | `f.following` | Users who follow the target | ✅ |
| `getFollowing()` | `f.followers` | Users the target follows | ✅ |
| Trust score `$follower_count` | `f.following` | Incoming follower count | ✅ |

Swapping these (as this PR does) would invert the results.

### If lists appear swapped on the profile page

The bug may be in the frontend (e.g., rendering the followers API response under the "following" tab or vice versa), not in the API queries themselves.

PR #6377, opened 2026-03-27
Fix inverted follower/following logic in API and trust score calculation

### LINKED ISSUE #6299: Bug: Follower/Following Logic Swap in API and Trust Score Calculation

rachana192837:
## Summary
Fixes #6299

The `getFollowers()` and `getFollowing()` methods in `FollowersApiLoader` had 
their DQL join conditions swapped, causing the API to return completely inverted 
results — the followers list showed who the user follows, and the following list 
showed who follows them.

The same inversion existed in `TrustScoreCalculator`, where `getFollowerCount()` 
was counting outbound follows (people the user follows) instead of inbound 
followers (people who follow the user), silently corrupting trust scores for any 
user with an asymmetric follower/following count.

## Root Cause
The `UserFollowing` entity has two sides:
- `f.follower` → the user who **initiated** the follow
- `f.following` → the user who **is being followed**

All three affected queries had these two fields used in the wrong direction.

## Changes
**`src/Api/Services/Followers/FollowersApiLoader.php`**
- `getFollowers()`: fixed join to correctly return users who follow the given user
- `getFollowing()`: fixed join to correctly return users the given user follows

**`src/Moderation/TrustScoreCalculator.php`**
- `getFollowerCount()`: fixed join to count inbound followers instead of outbound follows

## How to Test
1. Find a user profile with an asymmetric follower/following count
2. Call `GET /api/user/{username}/followers` — should return who follows them
3. Call `GET /api/user/{username}/following` — should return who they follow
4. Verify the profile page Followers and Following tabs match correctly

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -535,7 +535,21 @@ Then the ".result" element should contain "expected"
 - Infinite scroll / lazy loading
 - Any operation that shows a loading spinner
 
-#### 7. Debugging Flaky Tests
+#### 7. Use `should not exist` When JS Removes Elements from DOM
+
+When JS calls `element.remove()`, the element is gone from the DOM entirely. Use `should not exist`, NOT `should not be visible`:
+
+```gherkin
+# WRONG — element is removed, not just hidden; isVisible() can't find it
+Then the element ".my-banner" should not be visible
+
+# CORRECT — element.remove() means it no longer exists in the DOM
+Then the element ".my-banner" should not exist
+```
+
+Rule: if the JS code calls `.remove()` on an element, assert with `should not exist`. Only use `should not be visible` when the element stays in the DOM but is hidden via CSS (`display: none`, `visibility: hidden`, etc.).
+
+#### 8. Debugging Flaky Tests
 
 If a test passes locally but fails in CI:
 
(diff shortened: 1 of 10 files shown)


rachana192837:
@dmetzner can u review the pr ? Would mean a lot.

PR #6420, opened 2026-03-28
COPPA/GDPR age gate, parental consent, parent portal, and minor restrictions

dmetzner:
## Summary

- **Age gate at registration**: date of birth required, minimum age 3
- **Parental consent for under-14**: satisfies both COPPA (13) and Austrian DSG §4 (14)
- **Parent email validation**: must differ from user's own email (case-insensitive)
- **Consent email sent after verification**: prevents spam to parent if child never verifies
- **Minor restrictions**: no comments (read or write), no follow/followers, not-for-kids content filtered
- **Reactions (likes) allowed** for minors — low-risk engagement
- **Email system redesign**: table-based layout, inline styles, absolute URLs, bulletproof buttons — works in Gmail/Outlook/Apple Mail
- **Mailpit added to Docker** for local email testing (port 8025)
- **Auto-delete unverified accounts**: `catrobat:clean:unverified-users` command (30 days default)
- **Admin email tool redesign**: preview/send all email templates, clean card UI
- **Reset password page redesign**: MDC text fields matching login/registration style
- **Complete registration flow** for existing users (OAuth) missing date of birth

## Test plan

- [ ] Register as adult (14+): no parent email required, full access
- [ ] Register as child (under 14): parent email required, parent email ≠ user email
- [ ] Verify email → consent email sent to parent (check Mailpit at localhost:8025)
- [ ] Parent clicks consent link → account unlocked
- [ ] Minor cannot see/post comments on any project
- [ ] Minor cannot see followers/following tabs on own or other profiles
- [ ] Minor cannot see "not for kids" projects in search/categories
- [ ] Admin email page: preview all 4 templates, send works, dropdown toggles fields
- [ ] Reset password flow: request → email → click link → new password → login
- [ ] Run `bin/console catrobat:clean:unverified-users --dry-run` to verify command

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dmetzner:
I found three blocking issues in the current implementation:

1. `src/Api/Services/User/UserApiProcessor.php:31-62`
`$this->user_manager->updateUser($user)` flushes the insert before `date_of_birth`, `is_minor`, `consent_status`, and `parent_email` are assigned. Because there is no second flush here, those new fields never reach the database, and `postPersist` has already fired before `consent_status`/`parent_email` are set. In practice, under-13 registrations will be stored with the migration defaults (`is_minor = false`, `consent_status = 'not_required'`, `parent_email = null`) and no parental-consent mail will be sent.

2. `src/Project/ProjectManager.php:107-110`
The new not-for-kids restriction only applies when code goes through `findProjectIfVisibleToCurrentUser()`, but several project API paths still bypass that helper entirely. `ProjectsApiLoader::findProjectByID()` and `getProjectsFromCategory()` still return raw repository results, and `ReactionsApiLoader::findProjectIfVisibleToCurrentUser()` reimplements visibility without the new minor check. That means minor accounts can still see restricted projects via `/api/project/{id}`, category feeds, and reactions endpoints even though the web controller path is blocked.

3. `src/Security/AccountStateEventListener.php:19-23`
The pending-consent gate is added below the existing exemptions, but `PUT /api/user` is still exempt. So an under-13 account with `consent_status = 'pending'` can continue performing profile writes (username/email/password/avatar/about) even though the PR summary says all write API operations should be blocked until parental consent is granted.

Happy to re-review once those are addressed.

Issue #6463, opened 2026-03-29
Code statistics: align CT scoring with rubric and make scores explainable


dmetzner:
## Summary
The recent code statistics rework already solved the old backend/storage problem by persisting `project_code_statistics`, generating stats on upload, and backfilling on demand (#6433, merged March 28, 2026; #6443, merged March 29, 2026).

I would **not** reopen the old Jira-style ticket as-is.

The biggest remaining gap is now **score quality and explainability**: the current implementation stores useful counts, but the displayed computational thinking (CT) scores are still heuristic raw counts rather than the rubric-based levels described in the older stakeholder material.

## Why a follow-up is still worth doing
I compared the current implementation with the old CT spreadsheet (`CT Statistics.xlsx`, dated May 19, 2021) and the shipped code.

### What is already done well
- Persistent DB model exists: `project_code_statistics`
- Upload-time generation exists via `ProjectAfterInsertEvent`
- On-demand generation for old projects exists
- The old parser was left untouched
- Parser unit tests exist and pass (`bin/phpunit tests/PhpUnit/Project/CodeStatistics/`)

### What still looks off
1. The current scores are **unbounded raw keyword counts**, not rubric levels.
   - The parser sums matching brick/script types directly in `CodeStatisticsParser::computeScores()`.
   - Example: abstraction can become `5`, `8`, `12`, etc. depending on brick count.
   - The older rubric is threshold-based by concept/level (`Basic`, `Developing`, `Proficiency`), not raw occurrence totals.

2. Several rubric rules from the spreadsheet are not representable with the current parser.
   Examples:
   - Parallelism basic: `more than one StartScript`
   - Parallelism proficiency: same broadcast message / same background / same sensor / same variable triggers
   - Logical thinking proficiency: logical operators (`and`, `or`, `not`) inside formulas
   - Data representation basic: motion/look/visibility bricks such as `PlaceAtBrick`, `SetXBrick`, `SetYBrick`, `MoveNStepsBrick`
   The current parser only extracts type counts from `<script type="...">` and `<brick type="...">`, so it cannot reliably evaluate relationship-based or formula-based rules.

3. The frontend currently adds a **hardcoded bonus of +5** even when all scores are zero.
   - `assets/Project/CodeStatisticsInline.js` uses `const BONUS_POINTS = 5`
   - The Behat suite currently expects a total score of `5` for a project with no persisted scores and also for projects where all category scores are zero
[...]

PR #6465, opened 2026-03-29
Refresh code statistics with rubric scoring

dmetzner:
## Summary
- replace the legacy keyword-count CT scoring with a rubric-style evaluator that maps the 2021 stakeholder spreadsheet into basic/developing/proficiency checks
- persist `score_bonus`, `score_total`, and `scoring_version`, plus add a migration that marks existing rows as `legacy_keyword_counts_v1`
- refresh outdated persisted rows on read by reparsing current `code.xml` and exposing the upgraded scores through the API and inline project-page widget
- show backend-computed totals in the inline UI and render a real bonus row only when bonus points are actually earned
- expand parser/service coverage and tighten the code-statistics Behat fixtures around exact response values

## Why
Issue #6463 called out that the current implementation solved caching/storage, but not score quality or explainability. The old parser path was effectively raw keyword accumulation, the frontend invented `+5` bonus points on its own, and there was no way to distinguish legacy persisted rows from rubric-backed ones.

This PR turns the draft into a real follow-up slice: persisted statistics can now tell you which scoring model produced them, legacy rows can be upgraded lazily, and the UI/API are driven by stored backend totals instead of frontend guesses.

## Impact
- CT category scores now follow a versioned rubric model with `0-6` ranges per category
- projects can earn explicit bonus points for physics and extension usage
- API consumers get `score_bonus`, `score_total`, and `scoring_version`
- older persisted rows are transparently refreshed to the new rubric when current project files are available
- Behat fixture setup now clears previous code-stat rows for the same project, which makes exact-value assertions deterministic across reruns

## Validation
- `php -l src/Project/CodeStatistics/CodeStatisticsParser.php`
- `php -l src/Project/CodeStatistics/CodeStatisticsService.php`
- `php -l src/DB/Entity/Project/ProjectCodeStatistics.php`
- `php -l src/Api/Services/Projects/ProjectsResponseManager.php`
- `php -l src/Api/OpenAPI/Server/Model/CodeStatisticsResponse.php`
- `php -l src/System/Testing/Behat/Context/DataFixturesContext.php`
- `php -l tests/PhpUnit/Project/CodeStatistics/CodeStatisticsParserTest.php`
- `php -l tests/PhpUnit/Project/CodeStatistics/CodeStatisticsServiceTest.php`
- `php -l migrations/2026/Version20260329220000.php`
- `yarn eslint assets/Project/CodeStatisticsInline.js --no-warn-ignored`
- `yarn stylelint assets/Project/CodeStatisticsInline.scss`
- `yarn prettier assets/Project/CodeStatisticsInline.js assets/Project/CodeStatisticsInline.scss --check`
- `php vendor/phpunit/phpunit/phpunit --configuration phpunit.xml.dist tests/PhpUnit/Project/CodeStatistics/`
- `php vendor/behat/behat/bin/behat tests/BehatFeatures/api/projects/GET_project_id_code_statistics/code_statistics.feature --format progress --no-interaction`
- `git diff --check`

## Notes
- The targeted API Behat feature exercised cleanly with all scenarios progressing green, but the local Behat process did not return promptly after finishing in this environment, so I stopped there instead of waiting on teardown indefinitely.
- The browser-side Behat feature was not reliable enough to use as a gate from this isolated worktree; the parser/service PHPUnit slice and API Behat run provided the main verification signal here.

### DIFF
--- a/assets/Project/CodeStatisticsInline.js
+++ b/assets/Project/CodeStatisticsInline.js
@@ -9,17 +9,35 @@ const CHARACTERS = [
   { url: penguinSvgUrl, className: 'code-stats-animation--penguin' },
 ]
 
+const MAX_LEVELS = 3
+
 const SCORE_KEYS = [
-  { key: 'score_abstraction', transAttr: 'transAbstraction' },
-  { key: 'score_parallelism', transAttr: 'transParallelism' },
-  { key: 'score_logical_thinking', transAttr: 'transLogicalThinking' },
-  { key: 'score_synchronization', transAttr: 'transSynchronization' },
-  { key: 'score_flow_control', transAttr: 'transFlowControl' },
-  { key: 'score_user_interactivity', transAttr: 'transUserInteractivity' },
-  { key: 'score_data_representation', transAttr: 'transDataRepresentation' },
+  { key: 'score_abstraction', transAttr: 'transAbstraction', icon: 'extension' },
+  { key: 'score_parallelism', transAttr: 'transParallelism', icon: 'call_split' },
+  { key: 'score_logical_thinking', transAttr: 'transLogicalThinking', icon: 'psychology' },
+  { key: 'score_synchronization', transAttr: 'transSynchronization', icon: 'sync' },
+  { key: 'score_flow_control', transAttr: 'transFlowControl', icon: 'loop' },
+  { key: 'score_user_interactivity', transAttr: 'transUserInteractivity', icon: 'touch_app' },
+  { key: 'score_data_representation', transAttr: 'transDataRepresentation', icon: 'storage' },
 ]
 
-const BONUS_POINTS = 5
+function normalizeScoreValue(value) {
+  const numericValue = Number.parseInt(String(value ?? 0), 10)
+
+  return Number.isNaN(numericValue) ? 0 : numericValue
+}
+
+/**
+ * Map a 0–6 rubric score to a level count (0–3).
+ * basic = 1 pt, developing = 2 pts, proficiency = 3 pts.
+ * Thresholds: >= 1 -> level 1, >= 3 -> level 2, >= 6 -> level 3.
+ */
+function scoreToLevelCount(score) {
+  if (score >= 6) return 3
+  if (score >= 3) return 2
+  if (score >= 1) return 1
+  return 0
+}
 
 function animateNumber(el, from, to, duration) {
   if (from === to) {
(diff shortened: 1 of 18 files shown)


dmetzner:
## Code Review

Thanks for this thorough rework of the scoring system! The rubric-based approach is a significant improvement over keyword counting. A few issues to address:

**1. Generated model was hand-edited — will be overwritten**
`src/Api/OpenAPI/Server/Model/CodeStatisticsResponse.php` was modified directly instead of updating the OpenAPI spec and regenerating. The next `yarn run generate-api` will overwrite these changes. The spec changes look correct (new `score_bonus`, `score_total`, `scoring_version` fields), but the model file needs to be regenerated from them rather than hand-edited. Run `yarn run generate-api && bin/php-cs-fixer fix src/Api/OpenAPI/Server/ --using-cache=no`.

**2. `score_total` is computed in entity but has no DB column**
`ProjectCodeStatistics::getScoreTotal()` computes the total by summing fields at runtime. But the API response model has a `score_total` field (suggesting it's persisted). This works functionally but is confusing. Either persist it (add column for queryability — e.g. sorting projects by CT score) or document that it's a computed getter.

**3. `CodeStatisticsService` bypasses entity relationship for fresh query**
Changed from `$project->getLatestCodeStatistics()` to a direct `findOneBy()`. This works but breaks the entity relationship pattern. A comment explaining why (identity map caching after persist?) would help.

**4. Migration default `scoring_version = 'rubric_2021_v1'` for new rows**
The UPDATE sets existing rows to `legacy_keyword_counts_v1` ✓, but the column default means manually inserted rows get `rubric_2021_v1` without actually being rubric-scored. Consider `NULL` or `legacy_keyword_counts_v1` as the column default.

**5. `sumRubricLevels` scoring model (1+2+3=6) — document the additive model**
The method sums independent boolean levels rather than picking the highest. This means max score per category is 6, not 3. This is non-obvious — a brief docblock explaining "scores are additive: basic=1pt, developing=2pt, proficiency=3pt, max=6 per category" would help future contributors.

**6. Parser is no longer just regex-based**
`buildRubricContext()` uses `simplexml_load_string` + XPath. The class docblock still says "lightweight regex-based" — should be updated.

**7. Minor: CSS class rename `code-stats-row-bonus` → `code-stats-row--bonus`**
Good BEM convention. Verify no Behat tests reference the old class name.

Overall the rubric model is well-structured and the version tracking for lazy migration is a smart approach. Main thing to fix is #1 (regenerate instead of hand-edit).

